PR #5, opened 2019-11-29
Python2.7 compatibility

Timoses:


Timoses:
@mdellweg Do you know the source of above error? 
I'm not very familiar with the VCR fixture matching..

plugins/module_utils/pulp_helper.py
```diff
@@ -225,15 +228,16 @@ def chunked_upload(self, path, sha256, size):
...
-                            file=chunk_file_name,
-                            content_range=content_range,
-                        )
+                    temp_dir = mkdtemp(dir="/tmp")
+                    chunk_file_name = os.path.join(temp_dir, 'chunk.bin')
+                    with open(chunk_file_name, 'wb') as chunk_file:
+                        chunk_file.write(chunk)
+                    upload = self.uploads_api.update(
+                        upload_href=upload.pulp_href,
+                        file=chunk_file_name,
```
mdellweg: I think we should wrap this in a try finally construct. Also it might be valuable to look, whether ansible provides something there.

Timoses: > Thank you for this first of all contributions. I haven't thought, python 2 was still that important...
> This is not a full review yet, but what I can say already is that we need to find a way to run those tests in py2 also. Do you want to have a go there?

I guess when it comes to a lot of OS releases their stable repositories still use python2(.7) a lot.
How about adding python2.7 to the build pipeline? (https://docs.travis-ci.com/user/languages/python/#specifying-python-versions). I went ahead and added a commit. Results seem fine: https://travis-ci.com/mdellweg/ansible_modules_pulp/builds/138780423
Locally, it will be built using the available `python` executable. I think you can change it by using e.g. `python2.7 -m pytest`.

Timoses:
> Thank you for this first of all contributions. I haven't thought, python 2 was still that important...
> This is not a full review yet, but what I can say already is that we need to find a way to run those tests in py2 also. Do you want to have a go there?

I guess when it comes to a lot of OS releases their stable repositories still use python2(.7) a lot.
How about adding python2.7 to the build pipeline? (https://docs.travis-ci.com/user/languages/python/#specifying-python-versions). I went ahead and added a commit. Results seem fine: https://travis-ci.com/mdellweg/ansible_modules_pulp/builds/138780423
Locally, it will be built using the available python executable. I think you can change it by using e.g. python2.7 -m pytest.

plugins/module_utils/pulp_helper.py
Timoses: Do you mean the temporary directory? I'll add a commit.

mdellweg: Oh yes. Sorry this mobile app seems to be bad at adding comments to the right lines.

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: python
 python:
-  - 3.6
+  - "2.7"
```
mdellweg: The tests 'just' passed... Looks like we are lucky.
This is a good sign, as i remember that we had some strange problems with unstable list ordering in different python versions...

plugins/module_utils/pulp_helper.py
```diff
@@ -3,9 +3,13 @@
...
 # GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
 
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
 import os
+from shutil import rmtree
 import traceback
-from tempfile import TemporaryDirectory
+from tempfile import mkdtemp
```
mdellweg: I have not found an ansible provided solution for this, so ACK.

plugins/module_utils/pulp_helper.py
```diff
@@ -225,7 +231,8 @@ def chunked_upload(self, path, sha256, size):
                         end=offset + actual_chunk_size - 1,
                         size=size,
                     )
-                    with TemporaryDirectory() as temp_dir:
+                    try:
+                        temp_dir = mkdtemp(dir="/tmp")
```
mdellweg: I think you need to switch those two lines again.
There is no point in `rmtree(temp_dir)` if `mkdtemp` threw the exception.

plugins/modules/pulp_artifact.py
```diff
@@ -111,7 +113,7 @@ def main():
             sha256 = file_sha256
 
     if sha256 is None and module.params['state'] == 'absent':
-        module.fail_json(msg="One of 'file' and 'sha256' is required if 'state' is 'abesent'.")
+        module.fail_json(msg="One of 'file' and 'sha256' is required if 'state' is 'absent'.")
```
mdellweg: BTW: This is where i meant to put :+1:

plugins/module_utils/pulp_helper.py
Timoses: Good call.

Timoses:
Thanks for the review!

I've amended the last commit to set `temp_dir = mkdtemp(dir="/tmp")` before the `try:` statement.

PR #6, opened 2019-11-30
Add proxy_url and tls_validation option to file remote

Timoses:


tests/playbooks/file_sync.yaml
```diff
@@ -25,6 +25,8 @@
       pulp_file_remote:
         name: file_sync_test_file_remote
         url: "{{ pulp_fixtures_url }}/file/PULP_MANIFEST"
+        proxy_url: "{{ pulp_proxy_url | default(omit) }}"
+        tls_validation: "{{ pulp_tls_validation | default(omit) }}"
```
mdellweg: ...and here

tests/playbooks/file_remote.yaml
```diff
@@ -25,6 +25,8 @@
       pulp_file_remote:
         name: test_file_remote
         url: "{{ pulp_fixtures_url }}/file/PULP_MANIFEST"
+        proxy_url: "{{ pulp_proxy_url | default(omit) }}"
+        tls_validation: "{{ pulp_tls_validation | default(omit) }}"
```
mdellweg: Now comes the hard part. In this file we want to test file_remotes. So it would be helpful to add real values here (not `omit`). And that should most definitely lead to the need to rerecord this test.
(This is probably the right time to document that procedure...)
In short, you need to configure your pulplift instance in `vars/server.yaml` and call `make record_file_remote` (in a virtualenv?!).

Timoses:
Indeed, the addition of `proxy_url` and `tls_validation` to the test playbooks seem awkward now.

Initially, I started testing against a live Pulp instance not having had the time to dig into how the test environment was set up. In our environment we require using a proxy and TLS certificates are also exchanged which is why I added these to the tests.
After inspecting the test setup I realized that VCRpy actually enables offline testing.

My suggestion would be to leave these two options out of the other tests (as you suggested).
I would add to the `assert` checks to test that the values are actually set (no omit).
Sounds good?

tests/playbooks/file_remote.yaml
Timoses: See general comment: https://github.com/mdellweg/ansible_modules_pulp/pull/6#issuecomment-561291464

mdellweg:
> Sounds good?

Sounds exactly right.

Timoses:
> > Sounds good?
> 
> Sounds exactly right.

Good. I think it would make sense to test this again at a later time. Currently, it seems that the pulplift created instance uses a version which has `tls_validation`. However, the `requirements.txt` in this project uses an older `pulp-file-client` library which uses `ssl_validation`.
Updating the `pulp-file-client` leads to some other issues though. I suppose it would be worth fixing those before testing/recording test playbooks here.

(Meanwhile, I've updated the commit to implement some tests regarding proxy_url and tls_validation)

mdellweg:
I should update the pulp_file client bindings then (I pinned them, because there has been some confusion with version numbers).

mdellweg:
Any progress on the test fixture recording?

Timoses:
> cc @Timoses I hope this helps with #6

It did indeed. Thanks!

Should I squash the rerecording together with the actual changes?

plugins/modules/pulp_file_remote.py
```diff
@@ -120,7 +130,7 @@ def main():
         'name': module.params['name'],
     }
     desired_attributes = {
```
mdellweg: Actually, I am quite pleased to see, that this is all it needs at the implementation side. ;)

tests/playbooks/file_remote.yaml
```diff
@@ -88,6 +94,7 @@
           - result.changed == false
           - result.file_remote.name == "test_file_remote"
           - result.file_remote.url == "{{ pulp_fixtures_url }}/file/PULP_MANIFEST"
+          - result.file_remote.proxy_url == "http://proxy.int:3128"
```
mdellweg:
```suggestion
          - result.file_remote.proxy_url == "http://proxy.int:3128"
          - result.file_remote.tls_validation == False
```

mdellweg:
Since i wanted to change two small things in the tests asserts, i took the liberty to merge this by hand. So this PR should be considered as merged not _just_ closed.

PR #7, opened 2019-11-30
Make preparations to implement more pulp plugins

Timoses:
This PR tries to make some preparations to make the code more modular in terms of pulp plugin implementations.

Suggestions for improvements are very welcome!

(Note that this PR is based on https://github.com/mdellweg/ansible_modules_pulp/pull/5)

### DIFF
--- a/plugins/module_utils/pulp_core.py
+++ b/plugins/module_utils/pulp_core.py
@@ -1,0 +1,153 @@
+# -*- coding: utf-8 -*-
+
+# copyright (c) 2019, Matthias Dellweg
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+import os
+from shutil import rmtree
+from tempfile import mkdtemp
+from time import sleep
+import traceback
+
+from ansible.module_utils.basic import missing_required_lib
+
+try:
+    from pulpcore.client import pulpcore
+    HAS_PULPCORE_CLIENT = True
+except ImportError:
+    pulpcore = None
+    HAS_PULPCORE_CLIENT = False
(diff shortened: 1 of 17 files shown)


mdellweg:
So this is quite a big architectural change. This does not mean it is bad, or we should not make it, but i need a little more thought about this. I thought, maybe we could go into a direction, where we have less code around managing the importing of python modules (that is the part i'm not really happy with).

Timoses:
> I thought, maybe we could go into a direction, where we have less code around managing the importing of python modules (that is the part i'm not really happy with).

Which imports exactly are you referring to? Anyhow, take your time to think about it. It's worth it at this point, imo.

mdellweg:
I am thinking of all those properties, that lazyload the apis which lazyload the client_bindings. This looks like a lot of code duplication, and i wanted to take a broader look at this, when a few modules are implemented, to see what can be reused and what cannot. So one way could be to take your solution and refactor from there.

Timoses:
> I am thinking of all those properties, that lazyload the apis which lazyload the client_bindings. This looks like a lot of code duplication

Very true. I tried to follow that pattern since it seemed that at some points it was required to "translate" between the (self-written) API client and the actual pulp[core|-file]-client (as it was implemented already).
For example [here](https://github.com/mdellweg/ansible_modules_pulp/pull/7/files#diff-76c15490eb04ff33fae1ed1f12a910a8R62-R80) we don't just return the API but translate and eventually add more logic to make the API usage simpler from the client perspective.
So the indirection appears helpful.

All in all, I did not change much about the previous implementation other than cutting it into logical pieces and wrap them into classes (namely `PulpCoreApiClient`, `PulpFileApiClient` and `PulpEntityController`).

Timoses:
Rebase to latest master
+ updated `RETURN` section in `pulp_file_*` modules to reflect change.

Timoses:
This PR has become obsolete in the face of https://github.com/mdellweg/ansible_modules_pulp/pull/10

PR #10, opened 2020-02-09
Refactor helper

mdellweg:


mdellweg:
@Timoses This is what i came up with your idea of the PulpEntity.
What do you think? Can you easily implement more modules on this basis?

Timoses:
Actually responded to the wrong pull request, I see.
As said already: It looks great and I think it's a good approach to implementing more pulp plugins. Looking forward to give it a try once I have some more time again.

mdellweg:
Then i will merge this one as it sounds that we agree it is definitely an improvement and i already based the other work on top of it. ;)

PR #11, opened 2020-02-09
Prepare collection

mdellweg:


plugins/module_utils/pulp_helper.py
```diff
@@ -51,177 +46,195 @@ def __init__(self, argument_spec={}, **kwargs):
...
+        desired_attributes = self.desired_attributes.copy()
+        # drop 'file' because artifacts as well as content units are immutable anyway
+        desired_attributes.pop('file', None)
+        for key, value in desired_attributes.items():
+            if getattr(self.entity, key, None) != value:
+                setattr(self.entity, key, value)
+                changed = True
+        if changed:
+            if not hasattr(self.api, 'update'):
+                self.module.fail_json(msg="This entity is immutable.")
```
Timoses: Wouldn't it make sense to check this before changing `self.entity` (so at the entry point of the method like in `create` above)?

galaxy.yml
```diff
@@ -1,10 +1,11 @@
-namespace: "pulp"
-name: "pulp"
+namespace: "mdellweg"
+name: "squeezer"
 authors:
   - "Matthias Dellweg <2500@gmx.de>"
+  - "Timo Funke <timo.funke@biotronik.com>"
```
Timoses: <3
Could you change it to `timoses@msn.com` please?

Timoses:
Started skimming through the PR a bit. Looks great!! You were quite quick with it.
I think we're on a good path towards being able to implement other pulp plugins.
Haven't finished looking through it all, but will try to continue as soon as possible.

galaxy.yml
mdellweg:
```suggestion
  - "Timo Funke <timoses@msn.com>"
```

plugins/module_utils/pulp_helper.py
mdellweg: Actually, this logic isn'r even new...
But it fails only if an actual change is attempted for a reason. Even the module for an immutable entity should be idempotent. If this check was at the beginning, second call with `state=present` would fail instead of succeeding with `changed=false`.

mdellweg:
@Timoses 
It's surely not wron to look at this PR also.
It enables the thing to be used as an actual collection, restructures how the modules are brought in place for testing and adds a lot of noisy linting.

README.md
```diff
@@ -3,10 +3,26 @@
 ansible modules for Pulp aka Squeezer
 ===
 
-This project project aims to provide a complete set of ansible modules to control a pulp3 server in a descriptive way.
+This project project provides a set of ansible modules to control a pulp3 server in a descriptive way.
```
Timoses: "project project" (duplicate)

mdellweg:
```suggestion
This project provides a set of ansible modules to control a pulp3 server in a descriptive way.
```

Timoses:
> @Timoses
> It's surely not wron to look at this PR also.
> It enables the thing to be used as an actual collection, restructures how the modules are brought in place for testing and adds a lot of noisy linting.

Just spent a bit of time looking at Ansible Collections.
So I suppose its goal is to generalize the concept of roles to externalize implementation of all kinds of Ansible content into collections.
So in the end one could simply write in a requirements.yml
```
collections:
- name: squeezer
  src: https://github.com/mdellweg/ansible_modules_pulp
  scm: git
  version: master
```
??

I've looked through the code changes and so far didn't find anything that looked suspicuous ; ).

mdellweg:
Yes, that is roughly the point. The generalization is in the way, that a collection can contain roles, playbooks and plugins afaik.

PR #13, opened 2020-03-30
Change PulpAnsibleModule into a context manager

mdellweg:
This is my take on how the ansible module api should have been.

### DIFF
--- a/plugins/module_utils/pulp_helper.py
+++ b/plugins/module_utils/pulp_helper.py
@@ -33,27 +33,26 @@ CONTENT_CHUNK_SIZE = 512 * 1024  # 1/2 MB
 
 
 class PulpAnsibleModule(AnsibleModule):
-
-    def __init__(self, argument_spec=None, **kwargs):
-        if argument_spec is None:
-            argument_spec = {}
-        self._changed = False
-
-        spec = dict(
+    def __init__(self, **kwargs):
+        argument_spec = dict(
             pulp_url=dict(required=True),
             username=dict(required=True),
             password=dict(required=True, no_log=True),
             validate_certs=dict(type='bool', default=True),
         )
-        spec.update(argument_spec)
-        kwargs['supports_check_mode'] = kwargs.get('supports_check_mode', True)
-        super(PulpAnsibleModule, self).__init__(argument_spec=spec, **kwargs)
+        argument_spec.update(kwargs.pop('argument_spec', {}))
+        supports_check_mode = kwargs.pop('supports_check_mode', True)
+        super(PulpAnsibleModule, self).__init__(argument_spec=argument_spec, supports_check_mode=supports_check_mode, **kwargs)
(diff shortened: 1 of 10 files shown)


mdellweg:
@Timoses I'd like your :eyes: on this.

Timoses:
Looks interesting. The `argument_spec` stuff in Pulp(Entity)AnsibleModule is great, makes it much cleaner.

I kind of thought about making all the `PulpEntity` classes independent from the module.
I guess this is a start in that `fail_json` is no longer directly called from the Entities.

Perhaps it could even go a step further?
The Entity itself could keep track of `changed` and `results`.
So instead of `self.module.set_changed()` it would be `self._changed = True` and the module would check the entity afterwards and decide upon the entity's state how to continue (e.g. exit_json with changed, or fail_json, ...)

I'm not sure how feasible this is. Plus, how much sense it makes to disentangle PulpEntity from Ansible module logic. Afterall, it's currently used in Ansible Module code. So where would be the plus side? What's the goal?

In regard to using the Modules with `with`, I'm not sure what the upside is (other than making use of the `fail_json` upon a thrown exception from within the `with` statement). Also, doesn't this make it so that the module class **must** be used with `with`? As otherwise exceptions wouldn't correctly trigger a `fail_json` call in the module?

mdellweg:
> Looks interesting. The `argument_spec` stuff in Pulp(Entity)AnsibleModule is great, makes it much cleaner.
> 
> I kind of thought about making all the `PulpEntity` classes independent from the module.
> I guess this is a start in that `fail_json` is no longer directly called from the Entities.

I mostly did this, because i dislike the way, `fail_json` breaks out of the control flow. Exceptions are built for this. You can catch them selectively, refine them rethrow them or rescue your system.

> Perhaps it could even go a step further?
> The Entity itself could keep track of `changed` and `results`.
> So instead of `self.module.set_changed()` it would be `self._changed = True` and the module would check the entity afterwards and decide upon the entity's state how to continue (e.g. exit_json with changed, or fail_json, ...)
> 
> I'm not sure how feasible this is. Plus, how much sense it makes to disentangle PulpEntity from Ansible module logic. Afterall, it's currently used in Ansible Module code. So where would be the plus side? What's the goal?

Hmm, not immediately obvious to me. The whole bring this entity in that state is well tailored for ansible.

> In regard to using the Modules with `with`, I'm not sure what the upside is (other than automagically making use of the `fail_json` upon a thrown exception from within the `with` statement). Also, doesn't this make it so that the module class **must** be used with `with`? As otherwise exceptions wouldn't correctly trigger a `fail_json` call in the module?

That is right. But i do not see that as a big problem, as it is going to be common boilerplate code.
(`__exit__` is the closest thing in python to a destructor.)

Timoses:
Ok, I can agree with all that you've mentioned!

lgtm :+1:

PR #19, opened 2020-07-07
Update README.md to reflect upload to galaxy

mdellweg:


README.md
```diff
@@ -33,40 +31,55 @@ Available modules
...
-Those playbooks are meant to test one specific module, and are usually structured in three sections:
+There is usually one playbook per module that it is meant to test, but that is not a hard requirement.
+
+The playbooks are usually organized in a three-act structure:
 
- * In the first part, ran against `localhost`, fixtures can be set up.
- * The second part should contain the actual tests and is executed on the virtual host `tests` to allow recording.
- This usually involves calling the module in question several times with varying parameters and verifying it's output.
- * The third part, again for `localhost`, provides the the opportunity to undo part one.
+ * In the first act, the setup, ran against `localhost`, fixtures can be prepared.
```
melcorr, line 65: Hey @mdellweg 
I am going to ask a lot of questions. We can have a quick phone conversation to review if it is easier. 

this might be a really stupid question, but does this sentence mean the same thing: 

"In the first act, the setup runs against `localhost` to prepare the fixtures"? 

Also, does "fixtures" mean "a collection of raw fixture data, scripts for compiling that data into a useful format, and make targets for controlling the whole mess." The technical writer in me would suggest we link to where 'fixtures' is defined for the hopeless among us who are reading this for the first time.

README.md
```diff
@@ -33,40 +31,55 @@ Available modules
...
+ * The second act, the confrontation, should contain the actual tests.
```
melcorr, line 66: Is confrontation a technical term? It sounds a small bit dramatic to me. By "should" contain does that mean that at this point the user must have done some previous step to ensure they are there? 

Does this make as much sense or is there something implied that is lost:
`The second act contains the actual tests.` 

All these questions are as much for me to learn as they are for the docs.

README.md
```diff
@@ -33,40 +31,55 @@ Available modules
...
+   It is executed on the virtual host `tests` to allow recording the requests to the rest api with the corresponding responses.
```
melcorr, line 67:
```suggestion
   It is executed on the virtual host `tests` to allow for requests to the REST API and the corresponding responses to be recorded.
```

README.md
```diff
@@ -33,40 +31,55 @@ Available modules
...
+   This usually involves calling the module in question several times with varying parameters and verifying it's output.
```
melcorr, line 68:
```suggestion
   This usually involves calling the module in question several times with varying parameters and verifying its output.
```

README.md
```diff
@@ -33,40 +31,55 @@ Available modules
...
+ * The third act, the resolution, again with the target `localhost`, provides the the opportunity to undo part one.
 
-In the first and third parts, no recording takes place.
-During playback, only the part in the middle is executed.
+In the first and third acts, no recording takes place.
+During playback, only the act in the middle is executed.
+Please make sure, that this act can run independently from the others.
```
melcorr, line 73: This is more a question rather than a suggestion, but are there any specific steps they can do to ensure this? Or is it just something that someone using this would know?

README.md
mdellweg: I was a little carried away by: https://en.wikipedia.org/wiki/Three-act_structure Maybe not the best idea.
The part you started to discuss here is the section someone hopefully reads, once he has decided to get his hands dirty and wants to contribute more modules/features. So i am assuming a some knowledge of programming and ansible here.
I am trying to explain the structure and the mechanics of the tests. In the context of tests, i think fixtures is a _thing_: Some entries you put in the database so your tests have something to chew on.

README.md
mdellweg: "the Confrontation" see above... (but it kind of matched)
I think it was "should" because i describe new to-be-added tests. But i can switch to describe the test that _are there_.

README.md
mdellweg: This is more like things not to do. e.g. not depending on variables that are set in the first part, because it will not be executed in the playback case.

README.md
```diff
@@ -33,40 +31,55 @@ Available modules
...
+
+The playbooks are usually organized in a three-act structure:
 
- * In the first part, ran against `localhost`, fixtures can be set up.
- * The second part should contain the actual tests and is executed on the virtual host `tests` to allow recording.
- This usually involves calling the module in question several times with varying parameters and verifying it's output.
- * The third part, again for `localhost`, provides the the opportunity to undo part one.
+ * In the first act, the setup, ran against `localhost`, fixtures can be prepared.
+ * The second act, the confrontation, should contain the actual tests.
+   It is executed on the virtual host `tests` to allow for requests to the REST API and the corresponding responses to be recorded.
```
mdellweg, line 67:
```suggestion
   It is executed on the virtual host `tests` to allow for requests to the REST API and their corresponding responses to be recorded.
```

README.md
melcorr: My 2c on this: 

As a Classics graduate, I am delighted to see the narrative models of Aelius Donatus alive and well in the 21st century. My only concern would be that this introduces terminology that is inconsistent with the other Pulp documentation. The "acts" "confrontation", while accurate, requires the user to understand this dramatic framework, which is not necessary and might confuse them. When it comes to usability, writing documentation in a consistent manner across all parts of the product reduces the mental load for the user. They grow to anticipate the way the information is framed.

mdellweg: I think the official terminology used in the context of ansible goes like "A playbook contains a list of plays". I'll change it to that.
And if only it helped to get some good discussions started on those lines, it was totally worth it.

PR #22, opened 2020-07-21
Use OpenAPI parser instead of precompiled bindings

mdellweg:


dkliban:
Do you need to add the openapi parser to the requirements.txt?

This looks cool

mdellweg:
No, the parser is no external library, but comes with this pr as module_utils.
That way those modules do not have any requirements anymore.

dkliban (approved):

PR #28, opened 2020-08-07
Add repair module

mdellweg:


plugins/modules/repair.py
```diff
@@ -0,0 +1,105 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = r"""
+---
+module: pulp_repair
+short_description: Repair corruted artifacts in a repository version on a pulp server
```
Timoses: typo: `corruted`

plugins/modules/repair.py
```diff
@@ -0,0 +1,119 @@
...
+    returned: if available
+"""
+
+import re
+import traceback
+
+from ansible_collections.pulp.squeezer.plugins.module_utils.pulp_glue import PulpAnsibleModule
+
+try:
+    import pulp_glue.file.context
```
mdellweg: We are getting closer, but we still need to find a way to find all pulp_glue plugin context modules, it seems.
Funny that this very issue (needing plugin specifics for something theoretically generic) was the original reason I was unhappy with the original design of this module.

PR #29, opened 2020-08-13
Add compatibility with pulp 3.6 using openapiv3

mdellweg:


cognifloyd:
I just applied this to work with pulp 3.6. Much of it is working.

One issue. With the `python_remote` module, the valid options are `immediate`, `on-demand`, and `streamed`. But if I try to use `on-demand` I get 400 errors.

According to the [api doc](https://pulp-python.readthedocs.io/en/latest/restapi.html#operation/remotes_python_python_create), it should be `on_demand` instead of `on-demand`. Is this a side effect of the switch from openapi 2 to openapi 3 (ie something in scope for this issue)? Or was this incorrect in the modules to begin with (and therefore should be a separate issue/PR)?

mdellweg:
It's not covered by the tests, so that is hard to descide, but also irrelevant (whether it was a bug before).
If you want to fix it in a separate PR (ideally with an adjusted test) that would be great. Otherwise i might end up including the fix here tomorrow.
Thank you for finding it.

cognifloyd (approved):
LGTM

dkliban (approved):

PR #30, opened 2020-08-17
Fix on_demand download policy

mdellweg:
Reported by @cognifloyd here: https://github.com/pulp/squeezer/pull/29#issuecomment-674478506


cognifloyd (approved):
LGTM

dkliban (approved):

PR #32, opened 2020-08-17
Add RPM Modules

cognifloyd:
fixes: #7222
https://pulp.plan.io/issues/7222

This adds modules for configuring pulp_rpm.
This is a copy of the pulp_python modules and tests with `s/Python/Rpm/` and `s/python/rpm` (then switch shebang back to python). Then I adjusted URLs, paths, and IDs to use rpm fixtures and repos.


cognifloyd:
Oops. This should be a draft PR.

dkliban:
@cognifloyd Please add the following to the commit message.

fixes: #7222 
https://pulp.plan.io/issues/7222

cognifloyd:
@mdellweg How does the vcr recording work? What's the best way to get an env setup to record the tests?

Do you just use pulplift?

I'm going to keep hacking at this to see if I can get it going.

cognifloyd:
So, I used my test pulp machine and ran `make record_rpm_*` to make the vcr recordings. I don't see any domains or credentials in the fixtures, so I added them to the PR and worked my way through each playbook till I got it to pass.

I think this is done, or at least it's the minimal set we need to configure pulp_rpm. There are a lot of other endpoints though, so we might need more modules to cover those.

plugins/modules/rpm_distribution.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# copyright (c) 2019, Matthias Dellweg
```
cognifloyd: This is a copy of ansible_distribution.py that replaces python with rpm.
Do I need to do anything different here? (and in the other modules)

cognifloyd (commented):
Question on copyright notices

plugins/modules/rpm_distribution.py
mdellweg: sure:
```suggestion
# copyright (c) 2020, Jacob Floyd
```
There is another author field in the `DOCUMENTATION` stanza. Can you also add your name to galaxy.yml please?

mdellweg:
Seems like you figured it out. :fireworks: FWIW, yes, i use a "pulp3-sandox-*" instance in pulplift for testing and recording the modules. Out of curiosity: Is your pulp 3.5 or 3.6?
Thank you so much for this contribution, and particularly proving that it is possible to follow the documentation about recording within one day.

@dkliban The ticket management in ci is not used in this repo. But should the ticket be changed to an epic and split for individual modules?

mdellweg:
> So, I used my test pulp machine and ran `make record_rpm_*` to make the vcr recordings. I don't see any domains or credentials in the fixtures, so I added them to the PR and worked my way through each playbook till I got it to pass.

Yes, we are trying to prevent including sensitive data:
https://github.com/pulp/squeezer/blob/develop/tests/vcr_python_wrapper.py#L55
https://github.com/pulp/squeezer/blob/develop/tests/vcr_python_wrapper.py#L91

cognifloyd:
>  Is your pulp 3.5 or 3.6?

3.6

> should the ticket be changed to an epic and split for individual modules?

I don't think so. This PR brings rpm support up to parity with ansible and python support.
Adding modules for different content type seems to go beyond that, so perhaps a new issue for expanded functionality should be filed to put on #34.

mdellweg:
Honestly it has not yet been discussed, whether this repository needs to follow the ticket policy.
So for now i don't need a ticket on #34.

plugins/modules/rpm_distribution.py
```diff
@@ -0,0 +1,133 @@
...
+            key: module.params[key]
+            for key in ["base_path", "publication"]
+            if module.params[key] is not None
+        }
+
+        if content_guard_name is not None:
+            if content_guard_name:
+                content_guard = PulpContentGuard(module, {"name": content_guard_name})
+                content_guard.find()
+                if content_guard.entity is None:
```
mdellweg: That reminds me to fold this into `find` with a `failsafe` flag.

plugins/modules/rpm_repository.py
```diff
@@ -0,0 +1,95 @@
...
+def main():
+    with PulpEntityAnsibleModule(
+        argument_spec=dict(name=dict(), description=dict()),
+        required_if=[("state", "present", ["name"]), ("state", "absent", ["name"])],
+    ) as module:
+
+        natural_key = {"name": module.params["name"]}
+        desired_attributes = {}
+        if module.params["description"] is not None:
+            # In case of an empty string we try to nullify the description
```
mdellweg: I think this actually works nowadays.
```suggestion
            # In case of an empty string we nullify the description
```

tests/playbooks/rpm_remote.yaml
```diff
@@ -0,0 +1,128 @@
...
+        url: "http://mirror.linux.duke.edu/pub/centos/8/BaseOS/x86_64/os/"
+        proxy_url: "http://proxy.int:3128"
+        state: present
+      register: result
+    - name: Verify create remote
+      assert:
+        that:
+          - result.changed == true
+          - result.remote.name == "test_rpm_remote"
+          - result.remote.url == "http://mirror.linux.duke.edu/pub/centos/8/BaseOS/x86_64/os/"
```
mdellweg: You can put anything here, as the remote created by this test is never used.

cognifloyd:
I think this is ready to merge.

mdellweg (approved):
It's a big one, but i can't find anything suspicious. Good work!

plugins/modules/rpm_distribution.py
mdellweg: https://github.com/pulp/squeezer/pull/35

mdellweg:
If you like, you can rebase this on top of #35 (should make it easier)

cognifloyd:
Rebased on #35 (well, `develop` branch since it's merged) and addressed all your comments.

plugins/modules/rpm_distribution.py
```diff
@@ -0,0 +1,129 @@
...
+    type: dict
+    returned: when name is given
+"""
+
+
+from ansible_collections.pulp.squeezer.plugins.module_utils.pulp import (
+    PulpEntityAnsibleModule,
+    PulpRpmDistribution,
+    PulpContentGuard,
+    SqueezerException,
```
mdellweg, line 93: Not needed anymore.
Doesn't flake catch these?

mdellweg: hmm we don't use flake anymore.

mdellweg (approved):
I am going to merge it as is, as i seem to have missed the import on other modules as well.

PR #33, opened 2020-08-18
Reformat with black

cognifloyd:
[noissue]


mdellweg:
Why is this needed? My version of black (19.10b0 installed from pypi) does not see anything to change.

cognifloyd:
Ah. I had version 19.3b0. I just updated to 19.10b0 and it's no longer complaining.

PR #34, opened 2020-08-18
Pulp rpm package

cognifloyd:
This is built on top of #32 

It is not functional yet. I'm not sure how this should work. Unlike FileContent, you can't look up with `relative_path`.


mdellweg:
So what is the natural key for an RPM? nevra, digest? That would be the one to use.

daviddavis:
It's [nevra plus checksum (pkgId) and checksum_type](https://github.com/pulp/pulp_rpm/blob/5313249d28fc5a2ab3f34fdb4d6b0935cd4fcbdc/pulp_rpm/app/models/package.py#L260-L262).

mdellweg:
If the checksum is needed, it should also be sufficient (in this context).

mdellweg:
Can you rebase this? It were easier to look at it if it didn't contain all the other changes already merged.

keilr:
We need this module as well. Can I somehow help you with this PR?

ziegenberg:
We need this as well. Can someone pick this up?

cognifloyd:
Sorry. Since I wrote this, I changed jobs and haven't had time to pick it back up. Please run with what I've got. I won't have time to touch this for the foreseeable future.

ziegenberg:
I picked up your changes and based my PR #105 on them. Going forward from there.

ziegenberg:
I suggest closing this in favour of #105.

PR #37, opened 2020-08-20
Ansible collection remote

mdellweg:


dkliban (approved):
This looks good to me. +1 to keeping the two types of remotes in a single ansible module.

PR #42, opened 2020-09-17
Improve api caching

mdellweg:
* Only save api.json if parsing it seemd to work
* Add 'refresh_api_cache' option to all modules

### DIFF
--- a/plugins/doc_fragments/pulp.py
+++ b/plugins/doc_fragments/pulp.py
@@ -11,8 +11,6 @@ __metaclass__ = type
 class ModuleDocFragment(object):
     # Common pulp documentation fragment
     DOCUMENTATION = r"""
-requirements:
-  - pulpcore-client
 options:
   pulp_url:
     description:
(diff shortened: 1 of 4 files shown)


mdellweg:
@chouseknecht Would you mind verifying that this solves the problem you described?

plugins/doc_fragments/pulp.py
```diff
@@ -38,6 +36,12 @@ class ModuleDocFragment(object):
       - If no value is specified, the value of the environment variable C(SQUEEZER_VALIDATE_CERTS) will be used as a fallback.
     type: bool
     default: true
+  refresh_api_cache:
+    description:
+      - Whether the cached api documentation should be invalidated.
```
evgeni:
```suggestion
      - Whether the cached API documentation should be invalidated.
```

also maybe "specification" instead of "documentation"?

plugins/doc_fragments/pulp.py
```diff
@@ -38,6 +36,12 @@ class ModuleDocFragment(object):
...
+      - It is recomended to use this once with the M(status) module at the beginning of the playbook.
```
evgeni:
```suggestion
      - It is recommended to use this once with the M(pulp.squeezer.status) module at the beginning of the playbook.
```

(or whatever the correct FQCN is)

evgeni (commented):
makes sense to me, wording comments only

plugins/doc_fragments/pulp.py
mdellweg:
```suggestion
      - Whether the cached API specification should be invalidated.
```

evgeni (approved):

PR #43, opened 2020-10-01
Add live test mode

mdellweg:


Makefile
```diff
@@ -47,12 +49,18 @@ sanity: $(MANIFEST) | tests/playbooks/vars/server.yaml
 test: $(MANIFEST) | tests/playbooks/vars/server.yaml
 	$(PYTEST) $(TEST)
 
+livetest: $(MANIFEST) | tests/playbooks/vars/server.yaml
+	pytest -v 'tests/test_playbooks.py::test_playbook' --vcrmode live
```
mdellweg, line 53: Here the tests must be run sequentially.

tests/conftest.py
```diff
@@ -3,10 +3,15 @@
 
 def pytest_addoption(parser):
     parser.addoption(
-        "--record", action="store_true", help="record new server-responses"
+        "--vcrmode",
+        action="store",
+        default="replay",
+        help="mode for vcr recording; one of ['replay', 'record', 'live']",
```
evgeni, line 10: I think you can use `choices` here like in argparse (https://docs.python.org/3/library/argparse.html#choices, https://docs.pytest.org/en/latest/reference.html#pytest.config.argparsing.Parser.addoption), then you don't need the `assert` later.

evgeni (changes requested):
I like that! One suggestion, good to go after that.

tests/conftest.py
evgeni: Yes, I had to read `pytest` source to figure that out.

mdellweg: Now that i found the docs, i know that this parser is _from_ argparse. ;)

evgeni (approved):

PR #49, opened 2021-01-02
Reload api docs in status module

mdellweg:


dkliban (approved):

plugins/modules/status.py
```diff
@@ -47,10 +47,29 @@
 )
 
 
+def has_pulpcore_3_9(component_versions):
+    pulpcore_version = list(map(int, component_versions["pulpcore"].split(".")[:2]))
```
evgeni: Why not `LooseVersion` from `distutils`?

mdellweg: Seems to pass sanity instead of `packaging.version.parse`. :+1:

PR #53, opened 2021-02-26
Add more filtering options for Debian remotes

cityofships:
- architectures
- components


mdellweg (changes requested):
Thank you very much for this addition! And kudos for adding tests without me asking!

Quite honestly, it looks, like you did not follow the rerecord instructions: https://github.com/pulp/squeezer#testing
Those prerecorded fixtures are really not meant to be edited by hand.

cityofships:
That's true, I took a shortcut here. Let me fix this.

mdellweg (approved):

PR #60, opened 2021-04-29
Support for mirror option in sync tasks

vonsch:


vonsch:
Please note I tested only pulp_rpm codepath in our environment, but I'm 99% sure that the other code paths will work as well

plugins/module_utils/pulp.py
```diff
@@ -213,11 +213,15 @@ def delete(self):
...
         self.module.set_changed()
 
-    def sync(self, remote_href):
+    def sync(self, remote_href, mirror):
         if not hasattr(self, "_sync_id"):
             raise SqueezerException("This entity is not syncable.")
+
+        body = {"remote": remote_href}
+        if mirror:
+            body["mirror"] = mirror
```
mdellweg, line 222:
```suggestion
        if mirror is not None:
            body["mirror"] = mirror
```
If we also send mirror=False, we can overwrite (on demand) whatever the server sets as default.

plugins/modules/ansible_sync.py
```diff
@@ -35,6 +35,10 @@
       - Name of the repository
     type: str
     required: true
+  mirror:
+    description:
+      - If True, synchronization will run in "mirror" mode instead of "additive"
```
mdellweg, line 40: Maybe explain that mirror mode removes content that disappeared upstream.
Also False means additive (no content is removed).
If not specified the server default is used.
(see my other comment for this tri-state)

mdellweg (commented):
Thank you for this addition.
Would you mind to add a test for this, too?

vonsch:
Obsoleted by https://github.com/pulp/squeezer/pull/86

PR #64, opened 2021-07-02
doc: EXAMPLES replace `api_url` with `pulp_url`

vkvm:
The EXAMPLES docstring in all modules is out-of-date it seems. I've only started using a few of these modules, and I wasted time with the examples before I realized they were wrong. I expect others will too.

Changed:

- There is no longer an `api_url` parameter. Replaced with `pulp_url`.
- A scheme/protocol (http://) is needed. Added it.

Should I also change the port number?

### DIFF
--- a/plugins/modules/access_policy.py
+++ b/plugins/modules/access_policy.py
@@ -76,7 +76,7 @@ author:
 EXAMPLES = r"""
 - name: Dump all access policies
   access_policy:
-    api_url: localhost:24817
+    pulp_url: https://pulp.example.org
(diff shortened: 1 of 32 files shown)


mdellweg:
Yes, that is probably a good idea.
Let's change it to "https://pulp.example.org" to make it most instructive.

vkvm:
> Let's change it to "https://pulp.example.org" to make it most instructive.

Done

mdellweg (approved):

PR #69, opened 2021-07-25
Test nightly

mdellweg:


mdellweg:
Failure seems to be: https://github.com/pulp/pulpcore/pull/1519

It sometimes looks to me that pulp-cli and squeezer are our best canaries for nightly integration tests...

PR #71, opened 2021-09-17
Add modules for pulp_container plugin

markgoddard:
Adds the following modules:

* container_distribution
* container_remote
* container_repository
* container_sync


plugins/module_utils/pulp.py
```diff
@@ -1082,3 +1082,85 @@ def _href(self):
...
+    _create_id = "distributions_container_container_create"
+    _update_id = "distributions_container_container_update"
+    _partial_update_id = "distributions_container_container_partial_update"
+    _delete_id = "distributions_container_container_delete"
+
+    _name_singular = "distribution"
+    _name_plural = "distributions"
+
+    @property
+    def _href(self):
```
mdellweg, line 1102: Hmm, maybe we should stop maintaining this terrible backward compatibility hack. But that is a completely different topic. ;)

plugins/modules/container_distribution.py
```diff
@@ -0,0 +1,155 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = r"""
+---
+module: container_distribution
+short_description: Manage container distributions of a pulp api server instance
```
mdellweg, line 15: "of a pulp api server instance"
I never liked this piece. If you (as a fresh contributor) have a better idea to formulate this, bring it on.

plugins/modules/container_distribution.py
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# copyright (c) 2021, Mark Goddard
+# copyright (c) 2020, Jacob Floyd
```
mdellweg: Is this leftover copy and paste?
I don't think, we need it for the boiler plate. All the names will be in the collections galaxy.yaml file anyway.

plugins/modules/container_distribution.py
```diff
@@ -0,0 +1,155 @@
...
+      - Name of the content guard for the served content
+      - "Warning: This feature is not yet supported."
+    type: str
+    required: false
+extends_documentation_fragment:
+  - pulp.squeezer.pulp
+  - pulp.squeezer.pulp.entity_state
+author:
+  - Mark Goddard (@markgoddard)
+  - Jacob Floyd (@cognifloyd)
```
mdellweg: same here.

mdellweg (commented):
Thank you for this.
This all looks quite familiar (which is good). Nothing scary sticks out. However I want to ask you to add tests for these modules.

markgoddard:
> Thank you for this.
> This all looks quite familiar (which is good). Nothing scary sticks out. However I want to ask you to add tests for these modules.

It is mostly copypasta, but I did make a few adapatations based on the API ref, and I have tested it by hand. I saw the github actions fire up and realised that of course there should be automated tests to update. I'll take a look at those later.

mdellweg:
> > Thank you for this.
> > This all looks quite familiar (which is good). Nothing scary sticks out. However I want to ask you to add tests for these modules.
> 
> It is mostly copypasta, but I did make a few adapatations based on the API ref, and I have tested it by hand. I saw the github actions fire up and realised that of course there should be automated tests to update. I'll take a look at those later.

I know. With all the effort to bring the boilerplate down to a minumum, the really hard and tedious part is to provide and maintain reasonable test.

plugins/modules/container_distribution.py
markgoddard: how about "in a Pulp service"?

markgoddard: I can do that in a separate change for all the modules, once this one lands.

markgoddard:
Now with added tests. Sorry for the delay, other stuff got in the way.

Nice test suite - only criticism is that the VCR reproduces the lengthy polling of syncs :)

I think you'll need to approve me before the tests will run.

tests/playbooks/container_distribution.yaml
```diff
@@ -0,0 +1,143 @@
...
+      register: result
+    - name: Verify distribution of repository
+      assert:
+        that:
+          - result.changed == true
+          - result.distribution.name == "test_container_distribution"
+          - result.distribution.base_path == "test_container_base_path"
+          - ansible_check_mode or result.distribution.registry_path is regex(".*/test_container_base_path")
+          - result.distribution.repository == repository_result.repository.pulp_href
+          - ansible_check_mode or result.distribution.content_guard is regex("/pulp/api/v3/contentguards/*")
```
mdellweg, line 75: Maybe it's been too long and my brain wiped some corner cases, but why should these not work in check_mode?

tests/playbooks/container_sync.yaml
```diff
@@ -0,0 +1,90 @@
...
+        name: test_container_repository
+        state: absent
+    - name: Make repository present
+      pulp.squeezer.container_repository:
+        name: test_container_repository
+        state: present
+    - name: Make remote present
+      pulp.squeezer.container_remote:
+        name: test_container_remote
+        upstream_name: pulp/pulp-fixtures
```
mdellweg: Oh, i can see, where the sync time comes from. XD This is a really big container. Maybe take something much smaller?

Also other idea: We can try to filter all all answers for unfinished tasks in the VCR take. But that must be an optional flag, since there is at least one test for the task polling and canceling.

tests/playbooks/container_repository.yaml
```diff
@@ -0,0 +1,148 @@
...
+    - name: Remove description from repository (2nd try)
+      pulp.squeezer.container_repository:
+        name: test_container_repository
+        description: ""
+        state: present
+      register: result
+    - name: Verify remove description from repository (2nd try)
+      assert:
+        that:
+          - result.changed == false
```
mdellweg, line 126: This test may actually be superfluous. (Or let's say sufficient to be tested with file repos. It is mainly there, because with the python bindings for pulp it is impossible to remove the description.) OTOH it does not hurt. :shrug:

mdellweg (commented):
You nailed it! I am almost completely happy with the tests you produced. :1st_place_medal: 

Please reconsider the use of `ansible_check_mode` in the assertions.

mdellweg:
I forgot: Please run `black .` before committing.

tests/playbooks/container_distribution.yaml
markgoddard: The other fields can be determined from the module arguments, but these are only returned by Pulp once the distribution is created.

tests/playbooks/container_repository.yaml
markgoddard: I figured it was testing that it's possible to update an existing repository, both adding and removing a field? (and the second one is the standard idempotency follow up)

tests/playbooks/container_sync.yaml
markgoddard: pulp/test-fixture-1 seems to speed things up.

tests/playbooks/container_repository.yaml
mdellweg: Just keep it...

tests/playbooks/container_distribution.yaml
mdellweg: Right. `registry_path` is not `base_path`, i agree. But with respect to the content_guard, i disagree. See the corresponding `file_distribution` test.

https://github.com/pulp/squeezer/blob/develop/tests/playbooks/file_distribution.yaml#L86

markgoddard: It failed without it. This content guard is different, since the redirect gets added automatically.

mdellweg: Ah, I should have know that. :facepalm:

mdellweg (approved):

melcorr:
@markgoddard thanks so much for your contribution. If you'd like some SWAG, please complete this form: https://forms.gle/985j2A2kYsYJ9zZV6

markgoddard:
> @markgoddard thanks so much for your contribution. If you'd like some SWAG, please complete this form: https://forms.gle/985j2A2kYsYJ9zZV6

I tend to pass on SWAG, but it's certainly a nice gesture.

melcorr:
> > @markgoddard thanks so much for your contribution. If you'd like some SWAG, please complete this form: https://forms.gle/985j2A2kYsYJ9zZV6
> 
> I tend to pass on SWAG, but it's certainly a nice gesture.

No worries, the offer remains open if you change your mind.

markgoddard:
Hi @mdellweg, is there any plan to make a squeezer release?

mdellweg:
> Hi @mdellweg, is there any plan to make a squeezer release?

See #80

PR #72, opened 2021-09-24
Raise exception on missing artifact file

keilr:
`artifact` module:

- added "File not found." exception

### DIFF
--- a/plugins/modules/artifact.py
+++ b/plugins/modules/artifact.py
@@ -76,10 +76,11 @@ RETURN = r"""
     returned: when file or sha256 is given
 """
 
-
+import os
 from ansible_collections.pulp.squeezer.plugins.module_utils.pulp import (
     PulpEntityAnsibleModule,
     PulpArtifact,
+    SqueezerException,
 )
 
 
@@ -91,6 +92,8 @@ def main():
 
         sha256 = module.params["sha256"]
         if module.params["file"]:
+            if not os.path.exists(module.params["file"]):
+                raise SqueezerException("File not found.")
             file_sha256 = module.sha256(module.params["file"])
             if sha256:
                 if sha256 != file_sha256:


plugins/modules/artifact.py
```diff
@@ -91,6 +91,8 @@ def main():
 
         sha256 = module.params["sha256"]
         if module.params["file"]:
+            if not os.path.exists(module.params["file"]):
+                raise Exception("File not found.")
```
mdellweg: Can we raise a SqueezerException here. There is no need to show the stacktrace, as we already know what's going wrong.

keilr: Thanks for the hint, I wasn't aware of the custom exception.

mdellweg: np It's handled here: https://github.com/pulp/squeezer/blob/develop/plugins/module_utils/pulp.py#L79

mdellweg:
Can you squash your commits into one before we merge?

keilr:
> Can you squash your commits into one before we merge?

done

mdellweg (approved):

mdellweg:
I'm convinced, the ci failure is unrelated.

fyi @gerrod
https://github.com/pulp/squeezer/pull/72/checks?check_run_id=3775351157#step:9:1029
https://github.com/pulp/squeezer/pull/72/checks?check_run_id=3775351157#step:9:1068
https://github.com/pulp/squeezer/pull/72/checks?check_run_id=3775351157#step:9:1144

mdellweg:
Thank you for this contribution!

gerrod:
@mdellweg - thanks for the mention, but not entirely sure how you'd like me to use this information... suspect you may have tagged the wrong developer 🤪

mdellweg:
> @mdellweg - thanks for the mention, but not entirely sure how you'd like me to use this information... suspect you may have tagged the wrong developer zany_face

@gerrod Right. Sorry!

PR #73, opened 2021-09-27
add Ansible 2.12 and Python 3.9 to CI and simplify matrix definition

evgeni:
instead of defining a big matrix and then a ton of excludes, only define
a default matrix for Python 3.8 (the default), and then add jobs for the
other Pythons using include

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,38 +14,24 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ["2.7", "3.6", "3.7", "3.8"]
+        python:
+          - "3.8"
         ansible:
           - stable-2.8
           - stable-2.9
           - stable-2.10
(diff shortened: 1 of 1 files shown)


evgeni:
not sure what makes 2.7/3.6/3.7 with devel expected/required, but those don't exist anymore as devel doesn't work with these Python versions anymore.

mdellweg:
> not sure what makes 2.7/3.6/3.7 with devel expected/required, but those don't exist anymore as devel doesn't work with these Python versions anymore.

That is in the configuration for branch protection... I can adjust that.

mdellweg (approved):

evgeni:
that orphans test, WASN'T ME! ;)

mdellweg:
> that orphans test, WASN'T ME! ;)

That may be the new orphan cleanup. (It's in the live tests.)

PR #74, opened 2021-09-27
Set orphan protection time to zero

mdellweg:
This allows to clean artifacts between tests.


evgeni:
I have no idea what this is doing, but it fixes the tests? yay!

mdellweg:
It stops protecting orphans for a day after last use. Therefore we get the module to report what we expect.

Issue #76, opened 2021-10-06
Improve instructions for development


markgoddard:
There are some instructions in the README for development, however there were definitely some parts left as an exercise for the reader. Here's what I did:

```
# Install deps
python3 -m venv venv
source venv/bin/activate
make test-setup

# Modify run_container.sh to run a Pulp in one container in the background
sed -i 's/^trap/#trap/' tests/run_container.sh
```

I then used the following script to record:

```
#!/bin/bash

set -ex

for m in repository remote sync distribution; do
    sudo docker rm -f pulp || true
    sudo rm -rf tests/settings/certs
    ./tests/run_container.sh
    make record_container_$m
done
sudo docker rm -f pulp || true
```

And this to test:

```
#!/bin/bash

set -ex

for m in repository remote sync distribution; do
    make test_container_$m
done
```

Eventually I worked out that you could do this (if you don't change the run_container.sh script):

```
./tests/run_container.sh make livetest_container_repository
```

However it fails if you run it again because the pulp container still exists (need to `docker rm -f pulp`).

mdellweg:
Thank you for the feedback. I usually record against a vm. But this container based approach is def an interesting addition to the toolbox. Regarding removing the container, this is a bit of a problem. The `run_container.sh` is really designed to run in the CI, and we need the container there to live after the tests, so we can call docker logs in the after_failure stage.

PR #77, opened 2021-10-06
run_container.sh: Remove pulp container on exit

markgoddard:
Removing the container means we can run the script multiple times.

Related: #76

### DIFF
--- a/tests/run_container.sh
+++ b/tests/run_container.sh
@@ -41,8 +41,13 @@ fi
 # show pulpcore/plugin versions we're using
 curl -s http://localhost:8080/pulp/api/v3/status/ | jq ".versions"
 
+exit_handler() {
+    "${CONTAINER_RUNTIME}" logs pulp > ${BASEPATH}/pulp.log 2>&1
+    "${CONTAINER_RUNTIME}" rm -f pulp
+}
+
 # shellcheck disable=SC2064
-trap "${CONTAINER_RUNTIME} stop pulp" EXIT
+trap exit_handler EXIT
 
 # Set admin password
 "${CONTAINER_RUNTIME}" exec pulp pulpcore-manager reset-admin-password --password password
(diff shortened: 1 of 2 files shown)


tests/run_container.sh
```diff
@@ -42,7 +42,7 @@ fi
 curl -s http://localhost:8080/pulp/api/v3/status/ | jq ".versions"
 
 # shellcheck disable=SC2064
-trap "${CONTAINER_RUNTIME} stop pulp" EXIT
+trap "${CONTAINER_RUNTIME} rm -f pulp" EXIT
```
mdellweg: We keep the container intentionally to be able to extract server logs post mortem. See the after_failure stage. At some point i wanted to do something like this: https://github.com/pulp/pulp-cli/blob/develop/pytest_pulp_cli/__init__.py#L99

markgoddard: That makes sense. Another way to do that would be to run `docker logs pulp` in the exit handler and redirect to a file.

mdellweg: I put some thought into this, and came to the following conclusions:
1. I like to have the container around (even locally) for post mortem analysis.
2. We could destroy the container before attempting to start a new one.
3. We could reuse an existing container instead of starting a new one every time (+ better startup time, - more side effects).
4. We should give the container a different name. "pulp" is to generic and taken from the oci-images example. People will get confused when this script starts messing around with it.
5. We use "the same (tm)" [as in copy'n'paste] in pulp-cli ci. So let's try not to add any squeezer specific features / names.

As i said before this script was never invented to be used outside of ephemeral ci, so a need for these changes seems reasonable when we try to make it fit for developer use.

Braindump for container names:
* pulp-test (probably still too generic)
* pulp-volatile
* pulp-ephemeral
* pulp-one-off
* pulp-tooling-ci

cc @ggainey you recently started to work extensively with local containers.

mdellweg:
Hey, I have a counter proposal here: https://github.com/pulp/squeezer/pull/79
It will delete the container in normal use cases (`--rm` when at startup). But you can opt to keep it around (`KEEP_CONTAINER=1`). Also it will capture the container logs test by test, which will make it easier to see what went wrong server side. Does that serve your needs?

mdellweg:
@markgoddard Are all your goals met by the other PR, and we can close this one?

markgoddard:
Alternative approach in #79 works for me.

PR #78, opened 2021-10-06
README: add run_container.sh

markgoddard:


README.md
```diff
@@ -57,8 +57,17 @@ Also it should not depend on any of the variables defined in `tests/playbooks/va
...
 To perform codestyle linting and ansible sanity checks, run `make lint sanity`.
 
-To (re-)record tests, you first need to setup a pulp instance ([pulplift](https://github.com/pulp/pulplift) is recommended here).
-With it's connection details configured in `tests/playbooks/vars/server.yaml`, you can run `make record_<playbook_name>`.
+To (re-)record tests or run live tests, you first need to setup a pulp
+instance, with it's connection details configured in
+`tests/playbooks/vars/server.yaml`. The `tests/run_container.sh` script is
+provided and allows you to run a command with a [Pulp in
+one](https://pulpproject.org/pulp-in-one-container/) container active. It
+requires Docker or Podman to be installed. For example:
```
mdellweg: I think you should mention, that you still need to have the `server.yaml` point in the right direction (especially, if this method is described, after the user tried to fiddle around with it for a vm). Would you mind replacing the dead reference to pulplift with mentioning the vagrant boxes in pulp_installer?

Also in this repo I prefer the on sentence per line docs.

markgoddard: I've split out a new subheading for testing against a pulp in one container. Could you provide something similar for testing against a VM given you are familiar with it?

mdellweg:
I like, where this is going. Can you add a second section to "run tests against a vagrant VM"? I can help you with the instructions there. It should contain a reference to https://docs.pulpproject.org/pulp_installer/pulplift/ .

Also as you stated, these instructions are not so useful if `run_container` fails to run a second time. How about force deleting the container _before_ starting a new one?

markgoddard:
> I like, where this is going. Can you add a second section to "run tests against a vagrant VM"? I can help you with the instructions there. It should contain a reference to https://docs.pulpproject.org/pulp_installer/pulplift/ .

Could you provide something?

> 
> Also as you stated, these instructions are not so useful if `run_container` fails to run a second time. How about force deleting the container _before_ starting a new one?

Agreed, trying to resolve that in #77. Force delete could work, but see what you think about my proposed solution.

README.md
```diff
@@ -57,8 +57,24 @@ Also it should not depend on any of the variables defined in `tests/playbooks/va
...
+
+```
+./tests/run_container.sh make livetest
+```
+
+Or to record test fixtures for the `rpm_repository` test:
+
+```
+./tests/run_container.sh make record_rpm_repository
+```
```
mdellweg, line 98: We can add a note here, that the container will not survive the command, but you can provide the `KEEP_CONTAINER` environment variable.
Also by providing `IMAGE_TAG=3.11` you can run against an older version of pulp.

README.md
```diff
@@ -57,8 +57,24 @@ Also it should not depend on any of the variables defined in `tests/playbooks/va
 To run the tests, you can either call `make test`, or `make test_<playbook_name>` to only run a specific one.
 To perform codestyle linting and ansible sanity checks, run `make lint sanity`.
 
-To (re-)record tests, you first need to setup a pulp instance ([pulplift](https://github.com/pulp/pulplift) is recommended here).
-With it's connection details configured in `tests/playbooks/vars/server.yaml`, you can run `make record_<playbook_name>`.
+To (re-)record tests or run live tests, you first need to setup a pulp instance, with it's connection details configured in `tests/playbooks/vars/server.yaml`.
```
mdellweg: My idea here is to provide a section for each method. Maybe start with this:

```suggestion
To (re-)record tests or run live tests, you need a running pulp instance.
Two common ways to provide that development server are explained below.

!!! Warning
    Do not use a production instance, as the tests might perform destructive actions.

### Recording tests against a Pulplift Vagrant environment

A full vm installation of pulp can easily be achieved by using [pulplift](https://github.com/pulp/pulp_installer/blob/master/docs/pulplift.md).
It is recommended to use one of the sandbox installations.
When the vm is up, you need to configure its connection details in `tests/playbooks/vars/server.yaml`.
Squeezer tests can now be run with:

+```
make livetest
+```

The fixtures for the `file_remote` test can be recorded with:

+```
make record_file_remote
+```
```

README.md
```diff
@@ -57,8 +57,49 @@ Also it should not depend on any of the variables defined in `tests/playbooks/va
...
+
+!!! Warning
+    Do not use a production instance, as the tests might perform destructive actions.
+
+### Recording tests against a Pulplift Vagrant environment
+
+A full vm installation of pulp can easily be achieved by using [pulplift](https://github.com/pulp/pulp_installer/blob/master/docs/pulplift.md).
+It is recommended to use one of the sandbox installations.
+When the vm is up, you need to configure its connection details in `tests/playbooks/vars/server.yaml`.
+Squeezer tests can now be run with:
```
mdellweg:
```suggestion
For example, to run all tests live against this vm:
```

mdellweg (commented):
I know, i am suggesting on my own text at this stage. XD

After this, we are probably good to go.

mdellweg (approved):

PR #79, opened 2021-10-07
Restructure log capturing in live tests

mdellweg:
This will capture server logs from the container in livetests test by
test, and remove the ephemeral test container in run_container.sh by
default. Also capturing the container logs is moved from an after
failure stage into run_container.sh.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,6 @@
+import os
 import pytest
+import subprocess
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -14,3 +16,35 @@ def pytest_addoption(parser):
...
+    def pulp_container_log(vcrmode):
+        if vcrmode == "live":
+            with subprocess.Popen(
+                [
+                    os.environ["PULP_LOGGING"],
+                    "logs",
+                    "-f",
+                    "--tail",
+                    "0",
+                    "pulp-ephemeral",
```
markgoddard, line 33: Unsure of the exact mechanics here, but IIUC this will show all logs since the container started. Is that what you're looking for?

mdellweg: Nah, the `--tail 0` part will only start capturing logs right now.

markgoddard: Ah, got it.

markgoddard (approved):
Looks like a good approach

PR #82, opened 2021-11-02
deb_publication: support simple, structured and verbatim modes

markgoddard:
Verbatim publisher uses a separate API endpoint.


plugins/modules/deb_publication.py
```diff
@@ -83,6 +91,7 @@ def main():
         argument_spec=dict(
             repository=dict(),
             version=dict(type="int"),
+            mode=dict(default="simple", choices=["simple", "structured", "verbatim"]),
```
mdellweg: I think, you can combine simple and structured in the same publication, but maybe we don't realy need this...

cc @quba42

quba42: It needs to be possible to set both simple and structured for the same APT publication. It is a valid use case.

markgoddard: Ack, I'll add parameters for both.

How about verbatim? Should that just be a third parameter?

mdellweg: Yes, i think, and it should conflict with the other two.

quba42: Yes. There are two publishers "APT" and "Verbatim". Of these "APT" has two possible modes: "simple" and "structured". At least one of the two modes must be used. Therefore the full list of different types of possible publications is:
* simple APT
* structured APT
* simple and structured APT
* verbatim

If you add something like "simple and structured" to your list of choices, that should cover every possibility.

plugins/module_utils/pulp.py
```diff
@@ -717,6 +717,24 @@ def _href(self):
...
+    _create_id = "publications_deb_verbatim_create"
+    _delete_id = "publications_deb_verbatim_delete"
+
+    _name_singular = "publication"
+    _name_plural = "publications"
+
+    @property
+    def _href(self):
+        return (
+            "deb_publication_href"
```
mdellweg: That will obviously not work, as it is the same as in line 714. Either the id was different, or this api was overshadowed.

markgoddard: Honestly, I don't know exactly what this part is for. Is it Pulp v2 vs Pulp v3? Or the version of openapi itself?

mdellweg: Good question. It is a work around for older versions of pulp3, namely <=3.5, using an openapi schema version 2. The API got a lot more consistent when we switched to drf-spectacular (generating openapiv3) with 3.6.

see commit 955374c2d4c26efd30238b0cc113d9d7e1eeb44e

mdellweg: I'm about to say, we should ditch compatibility with that dinosaur.

markgoddard: Done for this class. I suggest doing it for others in a separate patch.

markgoddard:
Do I need to do anything else here?

plugins/modules/deb_publication.py
```diff
@@ -26,6 +26,22 @@
       - Version number to be published
     type: int
     required: false
+  simple:
+    description:
+      - Whether to use simple mode. If none of simple, structured or verbatim
+        are specified, then a simple publication will be created
```
mdellweg: I would leave a note that these are incompatible with verbatim (if we keep the options that way).

plugins/modules/deb_publication.py
```diff
@@ -26,6 +26,22 @@
...
+    type: bool
+    required: false
```
mdellweg: If we use "default: False" instead of "required: False", wouldn't it make the code much cleaner?
Maybe a bit confusing with implicit simple however...

mdellweg (commented):
I kind of dig the idea, that currently we have 4 useful publication modes, and we could make it a choice field. It feels more intuitive to the user.
But i want to leave the decision to @quba42 here.

plugins/modules/deb_publication.py
markgoddard: I don't think it accurately represents the default though. The default is True, unless you specify any mode options.

mdellweg: That's what i mean by 'confusing' and also what kind of pushes me to the choice option, where it's easy to say `default: "simple"`, and users reading the docs will immediately understand.

markgoddard: Are there likely to be further modes added? It could lead to a combinatorial explosion of options

markgoddard: Another option is to make a separate module for verbatim pubs.

mdellweg: I currently don't see a need for more modes.

quba42: So we currently have three options:
- One module, with three separate boolean options (simple, structured, verbatim) that have relatively complex relations. (Current draft implementation).
- Two modules (separate module for verbatim pubs). This would mirror the structure of the pulp_deb API exactly, however it has proven to be difficult to understand and document.
- One module with one choice option that would need to include all useful combinations as a distinct choice.

If it is between option one and two, I would go for the separate verbatim module, and the one to one relation with the API endpoints. However, I actually think the third option is best. I am not to worried about "combinatorial explosion". New publication modes in the future are not unthinkable, but are expected to be rare. If we do add new publication modes in the future, I would probably add another separate endpoint like verbatim, rather than adding even more complexity to the APT publisher. That way we would get exactly one additional mode, and not a "combinatorial explosion". Note that the APT publisher is probably misnamed, since all the pulp_deb publishers are APT publishers, duh! What I am currently calling the APT publisher, is really the "simple and/or structured publisher", but that is a mouthful.

Taking all this together my suggestion is something like:
```
mode=dict(default="structured", choices=["structured", "simple",  "simple and structured", "verbatim"]),
```
This makes it very clear what the choices are.

Note that I have suggested "structured" as the default. This is a separate issue, but I feel like structured is perhaps the better default, since it recreates the structure of the upstream repo that was synced, which is probably what a naive user would expect from the default behaviour. (https://en.wikipedia.org/wiki/Principle_of_least_astonishment)
"simple" is kind of a special use case, namely: "I want all packages under a fixed distribution-component combination, regardless of the upstream repos' original structure"

markgoddard: Thanks for the response. I'm happy to go with option 3. I agree that structured is a better default, although it would break backwards compatibility.

mdellweg: About structured. It currently does not publish uploaded packages in the way you'd expect, so what is the bigger surprise?

quba42: Not sure if whitespace in a choice field is desirable. If not, just make it "simple_and_structured" or something...

quba42: Ok, maybe leave the default as simple then. We can always leave that as a discussion for a later time.

quba42: Or make the default "simple and structured", then everyone is equally unhappy! (No seriously let's not mess with this as part of this change).

quba42 (dismissed):
Looks very clean now, which makes me think option 3 was the right choice!

plugins/modules/deb_publication.py
```diff
@@ -26,6 +26,12 @@
       - Version number to be published
     type: int
     required: false
+  mode:
+    description:
+      - Mode to use when publishing.
+    type: str
+    default: simple
+    choices: ["structured", "simple",  "simple_and_structured", "verbatim"]
```
mdellweg:
```suggestion
    choices: ["structured", "simple", "simple_and_structured", "verbatim"]
```

mdellweg (commented):
Format nit pick. But i agree with @quba42

mdellweg (approved):

PR #83, opened 2021-11-02
deb_sync: support setting mirror parameter

markgoddard:
Setting mirror=true seems to be necessary to avoid a "Paths are
duplicated" error in some cases.


mdellweg (approved):
Nice!

markgoddard:
Do I need to do anything else here?

mdellweg:
Sorry for taking so long. I think this is good. Rerunning CI now...

mdellweg:
Thanks for your patience!

markgoddard:
And thank you for reviewing

PR #85, opened 2021-11-10
Fix idempotency for publications

### LINKED ISSUE #84: No idempotency for publications when multiple exist
# Steps to reproduce

Create multiple publications for a single repository version.

```
pulp rpm publication create --repository foo
pulp rpm publication create --repository foo
```

Use the rpm_publication module to define the same publication:

```
- name: Ensure RPM publications are defined
  pulp.squeezer.rpm_publication:
    <credential params>
    repository: foo
    state: present
```

# Expected

No new publication is created.

# Actual

A new, third publication is created.

# Analysis

This happens due to the find() method on PulpEntity looking for count==1. It uses a limit of one, so only one result is returned, but the count still gives the full number of matching items, which in this case is 2.

markgoddard:
If there are multiple publications of a repository version, then the
rpm_publication module will create another, rather than keeping or
updating one of the existing ones.

This happens due to the find() method on PulpEntity looking for
count==1. It uses a limit of one, so only one result is returned, but
the count still gives the full number of matching items, which in this
case is 2.

 This change fixes the issue by failing when we cannot find a unique
 entity matching the requested parameters.

Fixes: #84

### DIFF
--- a/plugins/module_utils/pulp.py
+++ b/plugins/module_utils/pulp.py
@@ -163,6 +163,16 @@ class PulpEntity(object):
         search_result = self.module.pulp_api.call(self._list_id, parameters=parameters)
         if search_result["count"] == 1:
             self.entity = search_result["results"][0]
+        elif search_result["count"] > 1:
+            # While we limit to one result, the count may be more than one.
+            # This may happen for models with an insufficient or non-existent "natural key"
+            # e.g. for a publication.
+            raise SqueezerException(
+                "Found multiple matches for {entity_type} ({entity_key}).".format(
+                    entity_type=self._name_singular,
+                    entity_key=self.natural_key,
+                )
+            )
         elif failsafe:
             self.entity = None
         else:


plugins/module_utils/pulp.py
```diff
@@ -161,7 +161,9 @@ def find(self, failsafe=True, parameters=None):
         parameters["limit"] = 1
         parameters.update(self.natural_key)
         search_result = self.module.pulp_api.call(self._list_id, parameters=parameters)
-        if search_result["count"] == 1:
+        # While we limit to one result, the count may be more than one.
+        # This may happen if the parameters are not unique, e.g. for a publication.
+        if search_result["count"] >= 1:
```
mdellweg: I think we should fail if we find more than one item.

mdellweg (commented):
This is a problem, but i don't think it's safe to assume, that the first one found is the one you want.
The fact that there is no way to identify publications with the same key as you create them is the real problem here for idempotence.
But as your reproducer already shows: If you stay within the world of idempotence (ansible) you cannot produce that situation. So I'd opt for letting the module fail instead of creating a third publication.

plugins/module_utils/pulp.py
markgoddard: I wondered if you might say this.

mdellweg: At least we agree that creating the third publication is the worst behaviour.

markgoddard: I was 50/50 on which way to go with the solution tbh

plugins/module_utils/pulp.py
```diff
@@ -161,8 +161,17 @@ def find(self, failsafe=True, parameters=None):
         parameters["limit"] = 1
         parameters.update(self.natural_key)
         search_result = self.module.pulp_api.call(self._list_id, parameters=parameters)
+        # While we limit to one result, the count may be more than one.
+        # This may happen if the parameters are not unique, e.g. for a publication.
```
mdellweg: Can you move that comment to the elif clause, and state that this can happen for models with insufficient or not existing "natural key" (that is how it is called in Django) e.g. Publications.

plugins/module_utils/pulp.py
```diff
@@ -163,6 +163,16 @@ def find(self, failsafe=True, parameters=None):
         search_result = self.module.pulp_api.call(self._list_id, parameters=parameters)
         if search_result["count"] == 1:
             self.entity = search_result["results"][0]
+        elif search_result["count"] > 1:
+            # While we limit to one result, the count may be more than one.
+            # This may happen for models with an insufficient or non-existent "natural key"
+            # e.g. for a publication.
```
mdellweg, line 169: :+1: thanks

mdellweg (approved):

PR #86, opened 2021-11-11
rpm_sync: support sync_policy

markgoddard:
sync_policy is a parameter for RPM syncs which allows switching to a
mirroring mode rather than additive. This can be useful particularly
when using on_demand remotes which may have removed content.

sync_policy was added in pulp_rpm 3.16.0, and replaces the deprecated
mirror parameter. If pulp_rpm is detected to be less than 3.16.0, this
change falls back to use the mirror parameter.


plugins/modules/rpm_sync.py
```diff
@@ -73,7 +108,20 @@ def main():
...
+            parameters = {"sync_policy": module.params["sync_policy"]}
+        elif module.params["sync_policy"] == "mirror_content_only":
+            raise SqueezerException(
+                "Cannot use sync policy 'mirror_content_only' with pulp_rpm<3.16"
+            )
+        else:
+            mirror = module.params["sync_policy"] == "mirror_complete"
+            parameters = {"mirror": mirror}
+
+        repository.process_sync(remote, parameters)
```
mdellweg: Sweet!

plugins/modules/rpm_sync.py
```diff
@@ -55,16 +64,42 @@
...
+
 from ansible_collections.pulp.squeezer.plugins.module_utils.pulp import (
     PulpAnsibleModule,
     PulpRpmRemote,
     PulpRpmRepository,
+    SqueezerException,
 )
 
 
+def _parse_version(version_str):
```
mdellweg, line 77: would using `distutils.LooseVersion` help here?

markgoddard: I think these days packaging.version.parse is preferred as distutils was never documented. However when I tried that I got complaints from the linter about the module import. I guess it's not part of the standard library.

plugins/modules/rpm_sync.py
```diff
@@ -73,7 +108,20 @@ def main():
         repository = PulpRpmRepository(module, {"name": module.params["repository"]})
         repository.find(failsafe=False)
 
-        repository.process_sync(remote)
+        # pulp_rpm supports sync_policy from 3.16.
+        # Earlier versions support only mirror.
+        rpm_version = module.pulp_api.api_spec["info"]["x-pulp-app-versions"]["rpm"]
```
mdellweg: Can you catch any error around this, and assume that versions not having it are definitely older than 3.16?

markgoddard: How about this:

```
rpm_version = module.pulp_api.api_spec.get("info", {}).get("x-pulp-app-versions", {}).get("rpm", ())
```

Which would result in an empty tuple, and always appear as a lower version?

mdellweg: sounds good.

markgoddard: done

mdellweg (approved):

markgoddard:
Thanks :)

PR #87, opened 2021-11-17
Private container distributions

markgoddard:
* Add a `private` parameter to the container_distribution module

### DIFF
--- a/plugins/modules/container_distribution.py
+++ b/plugins/modules/container_distribution.py
@@ -43,6 +43,11 @@ options:
       - "Warning: This feature is not yet supported."
     type: str
     required: false
+  private:
+    description:
+      - Whether to make the distribution private.
+    type: bool
+    required: false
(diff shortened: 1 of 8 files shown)


mdellweg (dismissed):

plugins/modules/container_distribution.py
```diff
@@ -109,6 +114,7 @@ def main():
             repository=dict(),
             version=dict(type="int"),
             content_guard=dict(),
+            private=dict(type='bool', default=False),
```
mdellweg: Actually, i would not default it here. If you leave it optional, then "not specifying it" equals "not touching it".

markgoddard: Makes sense.

mdellweg (approved):

markgoddard:
Ready to go?

PR #96, opened 2022-03-25
Use FQCN in examples

mdellweg:


plugins/modules/container_remote.py
```diff
@@ -75,11 +75,11 @@
 """
 
 RETURN = r"""
-  remotes:
+  pulp.squeezer.remotes:
```
evgeni: Return values have FQCNs now? ;)

mdellweg: :facepalm:

mdellweg: I would not risk a bet against it though.

mdellweg:
What happened to the hash filter in ansible 2.8? I swear it worked before...

plugins/modules/ansible_distribution.py
```diff
@@ -81,11 +81,11 @@
 """
 
 RETURN = r"""
-  distributions:
+  pulp.squeezer.distributions:
     description: List of ansible distributions
     type: list
     returned: when no name is given
-  distribution:
+  pulp.squeezer.distribution:
```
evgeni: no

evgeni (changes requested):
ansible_distribution.py is still fucked up ;)

evgeni (dismissed):

plugins/modules/status.py
```diff
@@ -24,7 +24,7 @@
 
 EXAMPLES = r"""
 - name: Read status from pulp api server
-  status:
+  pulp.squeezer.status:
     pulp_url: localhost:24817
```
evgeni, line 28: unrelated, but I think you could unify the `pulp_url`s ;)

evgeni (approved):

plugins/modules/status.py
mdellweg: Yes, i could.
https://github.com/pulp/squeezer/pull/98

PR #100, opened 2022-05-26
ostree support

sabre1041:
Support for ostree

### DIFF
--- a/plugins/module_utils/pulp.py
+++ b/plugins/module_utils/pulp.py
@@ -1203,3 +1203,89 @@ class PulpContainerRepository(PulpRepository):
             if self.module.pulp_api.openapi_version == 2
             else "container_container_repository_href"
         )
+
+
+# Ostree entities
+
+
+class PulpOstreeRepository(PulpRepository):
+    _list_id = "repositories_ostree_ostree_list"
+    _read_id = "repositories_ostree_ostree_read"
+    _create_id = "repositories_ostree_ostree_create"
+    _update_id = "repositories_ostree_ostree_update"
+    _partial_update_id = "repositories_ostree_ostree_partial_update"
+    _delete_id = "repositories_ostree_ostree_delete"
+    _sync_id = "repositories_ostree_ostree_sync"
+    _import_commits_id = "repositories_ostree_ostree_import_commits"
+
+    _name_singular = "repository"
+    _name_plural = "repositories"
+
+    @property
+    def _href(self):
+        return (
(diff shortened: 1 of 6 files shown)


mdellweg:
Great addition. I tend to want to ask for tests...

sabre1041:
@mdellweg I can look to add in tests, but it will take a little bit of time. So two options.

1. Integrate content as is. Create issue related to testing and assign to me
2. Wait until tests are added before integrating

Let me know which option that you would like to take

mdellweg:
I prefer option 2. This project (sadly) is very low traffic atm, and without some test coverage we can consider the addition broken from the time we merge it. Also i am in no rush here. Let me know if any of the testing process described in the readme is unclear.

sabre1041:
@mdellweg sounds good. I'll start tackling it soon and report back when its ready for a review

plugins/modules/ostree_distribution.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# copyright (c) 2019, Matthias Dellweg
```
mdellweg, line 4: I don't think, you should put my copyright there...
Also can you add yourself to the authors in galayx.yml?

plugins/modules/ostree_distribution.py
```diff
@@ -0,0 +1,142 @@
...
+    required: false
+  version:
+    description:
+      - Repository version number
+    type: str
+    required: false
+  content_guard:
+    description:
+      - Name of the content guard for the served content
+      - "Warning: This feature is not yet supported."
```
mdellweg, line 42: Is this really true?

plugins/module_utils/pulp.py
```diff
@@ -1203,3 +1203,89 @@ def _href(self):
...
+            "ostree_repository_href"
+            if self.module.pulp_api.openapi_version == 2
+            else "ostree_ostree_repository_href"
+        )
+
+    def import_commits(self, parameters=None):
+
+        repository_version = self.entity["latest_version_href"]
+
+        # In check_mode, assume nothing changed
```
mdellweg, line 1236: Is there a way to make this really idempotent? Maybe by checking if commits are part of the latest version? (Not so much of an ostree expert here.)

plugins/modules/ostree_distribution.py
sabre1041: Copy/paste error

PR #103, opened 2022-06-23
Add support for remote attribute in python_distribution

### LINKED ISSUE #102: Implement remote support for distributions
Currently `pulp.squeezer.python_distribution` does not support adding a remote. Therefore pull-through caching cannot be configured via squeezer.

`pulp.squeezer.python_distribution` should support the option `remote`.

ziegenberg:
This is to add the `remote` attribute support for `python_distribution` and fixes #102.


ziegenberg:
Tests are missing. Working on it.

plugins/modules/python_distribution.py
```diff
@@ -37,11 +37,17 @@
...
+    description:
+      - Name of the remote source to add to this publication
+    type: str
+    required: false
 extends_documentation_fragment:
   - pulp.squeezer.pulp
   - pulp.squeezer.pulp.entity_state
 author:
   - Matthias Dellweg (@mdellweg)
+  - Daniel Ziegenberg (@ziegenberg)
```
mdellweg, line 52: Please also add to galaxy.yml

ziegenberg: Fixed.

ziegenberg:
Hi @mdellweg, could you have a look at my PR? I guess, I'd need to modify fixtures, but I can't get tests to run on my workstation.

mdellweg:
Please take a look at the README.
Have you tried running `./tests/run_container.sh make livetest_python_distribution`? If sucessful, try `./tests/run_container.sh make record_python_distribution`.

ziegenberg:
Running `running ./tests/run_container.sh make livetest_python_distribution` fails with a terribly long error message.

See here: [error.log](https://github.com/pulp/squeezer/files/9009382/error.log)

ziegenberg:
I found SELinux in enforcing mode to be the reason for the failed container run.

ziegenberg:
I (re-)generated the fixtures.

plugins/modules/python_distribution.py
```diff
@@ -65,6 +71,16 @@
...
 
+- name: Create a python destribution with remote for pull-through caching
+  pulp.squeezer.python_distribution:
+    pulp_url: https://pulp.example.org
+    username: admin
+    password: password
+    name: new_python_distribution
+    base_path: new/python/dist
+    remote: new_remote
+    state: presetn
```
gerrod3:
```suggestion
    state: present
```

plugins/modules/python_distribution.py
```diff
@@ -37,11 +37,17 @@
       - "Warning: This feature is not yet supported."
     type: str
     required: false
+  remote:
+    description:
+      - Name of the remote source to add to this publication
```
gerrod3:
```suggestion
      - Name of the remote source to add to this distribution
```
Could also say "index".

plugins/modules/python_distribution.py
```diff
@@ -126,6 +145,12 @@ def main():
             else:
                 desired_attributes["content_guard"] = None
 
+        if remote_name is not None:
+            if remote_name:
```
gerrod3, line 151:
```suggestion
        if remote_name:
```
Both empty string `""` and `None` are falsey values.

ziegenberg: remove the extra if

plugins/modules/python_distribution.py
ziegenberg: Fixed.

plugins/modules/python_distribution.py
ziegenberg: Fixed the typo.

tests/playbooks/python_distribution.yaml
```diff
@@ -28,6 +28,10 @@
       pulp.squeezer.python_distribution:
         name: test_python_distribution
         state: absent
+    - name: Make distribution for pull-through absent
+      pulp.squeezer.python_distribution:
+        name: test_python_distribution_pull_through
+        state: absent
```
mdellweg, line 34: Can you copy this to the cleanup section at the end of this file?
I think, that is where you actually confused the `python_remote` test. (Still keep the more precise regex there...)

ziegenberg: Fixed and pushed.

plugins/modules/python_distribution.py
```diff
@@ -126,6 +145,11 @@ def main():
             else:
                 desired_attributes["content_guard"] = None
 
+        if remote_name is not None:
+            remote = PulpPythonRemote(module, {"name": remote_name})
+            remote.find(failsafe=False)
+            desired_attributes["remote"] = remote.href
```
mdellweg: Is it possible to remove the remote again from an existing distribution?
In the case of content guards (a few lines above), the extra if was there to determine whether `""` was provided to remove a configured guard.

gerrod3: Yes the remote can be removed.

ziegenberg: I'll change it accordingly, as soon as I'm in the office tomorrow.

ziegenberg: I amended my commit, adding the possibility to remove the remote and the necessary tests.

ziegenberg:
While I was at it, I also:

- changed it so we use FQCN for test tests in python_distribution.yml, now ansible-lint is happy
- added tests for removing content guards from python distribution

I put it into separate commits.

mdellweg:
Somehow the tests fail now. Can you maybe rerecord the fixtures for the python_distribution tests again? (Please check in all the newly recorded files.)

ziegenberg:
@mdellweg I rerecorded all fixtures.

Originally I didn't want to change test cases for unchanged tests. Didn't feel right to generate all test cases after a change and then use those test cases to validate those changes.

tests/playbooks/python_distribution.yaml
```diff
@@ -114,13 +124,31 @@
           - result.distribution.publication == publication_result.publication.pulp_href
           - result.distribution.content_guard is regex("/pulp/api/v3/contentguards/*")
 
+    - name: Remove content guart from distribution
```
mdellweg:
```suggestion
    - name: Remove content guard from distribution
```

tests/playbooks/python_distribution.yaml
```diff
@@ -131,7 +159,89 @@
...
+          - result.changed == true
+          - not result.distribution
+
+    - name: Delete distribution (2nd try)
+      pulp.squeezer.python_distribution:
+        name: test_python_distribution_pull_through
+        state: absent
+      register: result
+    - name: Verify delete distribution (2nd try)
+      ansible.builtin.assert:
```
mdellweg: I don't think we need those. Deletion of a distribution is already covered before.

mdellweg (commented):
Sorry, there are still some improvements to the tests i can see.
Also yes, the fixtures (for one test) need to be generated from the same run, as sometimes they refer to generated random ID's.

tests/playbooks/python_distribution.yaml
ziegenberg: This deletes a a distribution with a remote, the one before does not have a remote. But I can remove this test here.

tests/playbooks/python_distribution.yaml
ziegenberg: Fixed and pushed.

tests/playbooks/python_distribution.yaml
mdellweg: Yes, think of it that way: These tests should demonstrate that squeezer can talk to the api. It's not about testing the Pulp server. And the api call for deletion is simply the same. So if this can drop 2 seconds of every test run, i would prefer to have it. Also can you squash the commits in the end?

ziegenberg: Fixed the tests. And I squashed some of the commits. Do you want them to be squashed altogether? Cause they do different things.

mdellweg (approved):

mdellweg:
Thank you for digging through the test requirement.

PR #106, opened 2022-07-17
Rewrite with pulp-glue

mdellweg:
This is no more than a PoC at this stage, but the ultimate goal is to use the context framework of the pulp-cli in the ansible modules.


plugins/module_utils/pulp_glue.py
```diff
@@ -0,0 +1,217 @@
...
+            username=dict(required=True, fallback=(env_fallback, ["SQUEEZER_USERNAME"])),
+            password=dict(
+                required=True,
+                no_log=True,
+                fallback=(env_fallback, ["SQUEEZER_PASSWORD"]),
+            ),
+            validate_certs=dict(
+                type="bool",
+                default=True,
+                fallback=(env_fallback, ["SQUEEZER_VALIDATE_CERTS"]),
```
gerrod3, line 65: How does `fallback` work with a `default` value?

plugins/module_utils/pulp_glue.py
```diff
@@ -0,0 +1,217 @@
...
+            supports_check_mode=supports_check_mode,
+            **kwargs,
+        )
+
+        for import_error in import_errors:
+            if import_error[1] is not None:
+                self.fail_json(msg=missing_required_lib(import_error[0]), exception=import_error[1])
+
+        self.pulp_ctx = PulpSqueezerContext(
+            api_root="/pulp/",
```
gerrod3, line 86: Should `api_root` not be configurable?

plugins/modules/delete_orphans.py
```diff
@@ -38,15 +43,43 @@
...
+                item["message"].split(" ")[-1].lower(): item["total"]
+                for item in task["progress_reports"]
+            }
+        else:
+            # Fake it
+            summary = {
+                "artifacts": 0,
+                "content": 0,
+            }
+        module.set_changed()
```
gerrod3, line 82: Is there a way to have the`set_changed` only trigger for this step if orphan cleanup actually did work? Would that be in-line with ansible's idempotent philosophy?

plugins/modules/task.py
```diff
@@ -58,14 +58,55 @@
...
+                if not self.check_mode:
+                    if self.state == "canceled":
+                        self.context.cancel()
+                        # This should all be part of the cancel call...
+                        try:
+                            entity = self.pulp_ctx.wait_for_task(entity)
+                        except PulpException as e:
+                            if "canceled" in str(e):
+                                self.context.pulp_href = entity["pulp_href"]
+                                entity = self.context.entity
```
gerrod3: Is this a TODO for pulp-glue?

gerrod3 (commented):
I looked over the new modules and they all seem good to me. Left some extra non-blocking questions for understanding. Also, is this only for pulp_file, does that mean there will be as a large a change when you update it for the other plugins?

plugins/module_utils/pulp_glue.py
mdellweg: Yes, but not necessarily for feature parity...

plugins/modules/delete_orphans.py
mdellweg: I do not really know. This module always triggers a task / issues a POST|PUT|PATCH|DELETE request. So i consider it changing something.
Also when running in check_mode, there is no way to know that nothing would happen.
It is an action module after all not having a state parameter...

plugins/modules/task.py
mdellweg: Not anymore. But it's not yet released...

plugins/module_utils/pulp_glue.py
mdellweg: I can only guess... Probably set value, then env_fallback, then default.

PR #114, opened 2022-11-21
better rbac support in utils

Alex-Welsh:


tests/fixtures/ansible_distribution-1.yml
```diff
@@ -57,7 +57,7 @@ interactions:
       User-Agent:
       - Python-urllib/3.9
     method: GET
-    uri: https://pulp.example.org/pulp/api/v3/contentguards/certguard/x509/?limit=1&name=test_content_guard
+    uri: https://pulp.example.org/pulp/api/v3/contentguards/?limit=1&name=test_content_guard
```
mdellweg: Sorry, this is not the way to "adjust tests". Have you read about re-recording the pyvcr tapes in the README?

plugins/module_utils/pulp.py
```diff
@@ -538,20 +538,26 @@ def chunked_upload(cls, module, path, sha256, size):
...
     _name_plural = "content_guards"
 
     @property
     def _href(self):
-        return (
-            "x509_cert_guard_href"
-            if self.module.pulp_api.openapi_version == 2
-            else "certguard_x509_cert_guard_href"
-        )
+        return "r_b_a_c_content_guard_href"
```
mdellweg: Don't make this a property if it's not needed.
`_href = "r_b_a_c_..."`

mdellweg (changes requested):
Thank you for the improvements. Please read about rerecording the fixtures. Also i do not see the new content guard being used. Do you intend to write a module for it?

tests/fixtures/ansible_distribution-1.yml
Alex-Welsh: I had an issue yesterday that meant the files would just be deleted and not rewritten when recording. That seems to have resolved itself overnight

plugins/module_utils/pulp.py
Alex-Welsh: The intention was simply to allow `x_distribution` to use non-x509 content guards (in my case, I need RBAC content guards). I've removed the RBAC-specific stuff as it doesn't seem necessary at the moment.

tests/fixtures/ansible_distribution-1.yml
Alex-Welsh: Should be fixed now

mdellweg (approved):

PR #125, opened 2023-05-31
Add username and password to status.yaml

handflucht:
If username and password is not provided, the following error message is shown

``` 
["missing required arguments: username, password"](fatal: [somehost]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/bin/python"}, "changed": false, "msg": "missing required arguments: username, password"})
```

For that reason, password and username were added with the values defined in vars/server.yaml.


mdellweg:
This does not sound valid. The status endpoint works explicitely without authentication.

PR #134, opened 2023-10-31
Support the generate_repo_config property for rpm distributions

oyvindkolbu:
Since pulp_rpm version 3.23.0 it no longer generate `config.repo` files by default. Extend the module to support setting the property to generate them.

### DIFF
--- a/plugins/modules/rpm_distribution.py
+++ b/plugins/modules/rpm_distribution.py
@@ -37,6 +37,11 @@ options:
       - "Warning: This feature is not yet supported."
     type: str
     required: false
+  generate_repo_config:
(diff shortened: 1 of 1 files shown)


oyvindkolbu:
@mdellweg ?

mdellweg:
Sorry, forgot to answer. This needs tests to prevent regressions when refactoring the module.

netsandbox:
@oyvindkolbu maybe here should also be added a option `repo_config`, which was introduced by pulp_rpm version 3.24.0, to be able to modify the generated `config.repo`, because the default one is sometimes not suitable.

Issue #163, opened 2024-06-25
timeout connecting to pulp on K8S


maciej-markowski:
Hello,

I have deployed the pulp with use of the pulp operator on K8S cluster.
I have no problem connecting to the pulp and perform any CRUD operations with use of curl or pulp-cli, but when using squeezer, I am getting timeouts.
```
Traceback (most recent call last):
  File "/ansible/.ansible/tmp/ansible-tmp-1719302750.7030585-351-223232701389169/AnsiballZ_rpm_remote.py", line 107, in <module>
    _ansiballz_main()
  File "/ansible/.ansible/tmp/ansible-tmp-1719302750.7030585-351-223232701389169/AnsiballZ_rpm_remote.py", line 99, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File "/ansible/.ansible/tmp/ansible-tmp-1719302750.7030585-351-223232701389169/AnsiballZ_rpm_remote.py", line 47, in invoke_module
    runpy.run_module(mod_name='ansible_collections.pulp.squeezer.plugins.modules.rpm_remote', init_globals=dict(_module_fqn='ansible_collections.pulp.squeezer.plugins.modules.rpm_remote', _modlib_path=modlib_path),
  File "/usr/local/lib/python3.9/runpy.py", line 225, in run_module
    return _run_module_code(code, init_globals, run_name, mod_spec)
  File "/usr/local/lib/python3.9/runpy.py", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/usr/local/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/tmp/ansible_pulp.squeezer.rpm_remote_payload_sw7l1qdg/ansible_pulp.squeezer.rpm_remote_payload.zip/ansible_collections/pulp/squeezer/plugins/modules/rpm_remote.py", line 121, in <module>
  File "/tmp/ansible_pulp.squeezer.rpm_remote_payload_sw7l1qdg/ansible_pulp.squeezer.rpm_remote_payload.zip/ansible_collections/pulp/squeezer/plugins/modules/rpm_remote.py", line 84, in main
  File "/tmp/ansible_pulp.squeezer.rpm_remote_payload_sw7l1qdg/ansible_pulp.squeezer.rpm_remote_payload.zip/ansible_collections/pulp/squeezer/plugins/module_utils/pulp.py", line 72, in __enter__
[...]

PR #164, opened 2024-06-25
add pulp_timeout

maciej-markowski:
#163

this PR adds pulp_timeout parameter.

### DIFF
--- a/plugins/doc_fragments/pulp.py
+++ b/plugins/doc_fragments/pulp.py
@@ -42,6 +42,11 @@ options:
       - It is recommended to use this once with the M(pulp.squeezer.status) module at the beginning of the playbook.
     type: bool
     default: false
+  timeout:
+    description:
+      - Time in seconds to wait for tasks.
+    type: int
+    default: 10
 """
 
     GLUE = r"""
--- a/plugins/module_utils/openapi.py
+++ b/plugins/module_utils/openapi.py
@@ -40,6 +40,7 @@ class OpenAPI:
         password=None,
         validate_certs=True,
         refresh_cache=False,
+        timeout=10,
     ):
         self.doc_path = doc_path
 
@@ -60,6 +61,7 @@ class OpenAPI:
             headers=headers,
             validate_certs=validate_certs,
             force_basic_auth=True,
+            timeout=timeout,
         )
 
         self.load_api(refresh_cache=refresh_cache)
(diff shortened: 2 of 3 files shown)


plugins/doc_fragments/pulp.py
```diff
@@ -42,6 +42,11 @@ class ModuleDocFragment(object):
       - It is recommended to use this once with the M(pulp.squeezer.status) module at the beginning of the playbook.
     type: bool
     default: false
+  pulp_timeout:
+    description:
+      - timeout value for connecting to pulp api
```
mdellweg: This is not what timeout does.

maciej-markowski: `time in seconds to wait for successful connection to pulp api` - would that be a better description?

mdellweg: No, you are trying to make this parameter available to the not yet rewritten modules:
https://github.com/pulp/squeezer/blob/develop/plugins/doc_fragments/pulp.py#L63

maciej-markowski: I am not sure if I follow. 
I suppose what you're saying is that you're rewriting modules to use pulp_glue which already has this timeout?
As far as I can see, at least the rpm related modules are still using this openapi.py and it lacks this setting, which causes me some issues that I described in https://github.com/pulp/squeezer/issues/163.
Can you explain what I am supposed to do to get this merged?

mdellweg: Yes, basically.
You should use the same name and the same description at the very least. So rewriting the modules eventually will not break this again.

mdellweg (dismissed):

maciej-markowski:
I see the pipeline failed due to ansible removing support for python3.10
Can we still merge it and release or do we need to address this issues here?

mdellweg:
Nah, it would be better to fix the ci first.
Would you mind bumping some python versions in the workflows? Please make it a separate PR.

maciej-markowski:
> Nah, it would be better to fix the ci first. Would you mind bumping some python versions in the workflows? Please make it a separate PR.

#166

mdellweg:
Please rebase this.

maciej-markowski:
Are we going to merge https://github.com/pulp/squeezer/pull/166 first?

mdellweg:
I don't think, we need that anymore. I only adds more test combinations. And I am quite happy that we are down to 4 combinations, and still covering all supported ansible versions.

mdellweg (approved):

Issue #165, opened 2024-06-25
rpm_repository/retain_repo_versions


asphalter:
We'd like to specify how many repositories versions to keep online, but even if this param could be found out on pulp_cli, I can't find it either on squeezer.

Could you please verify and let me know?

Brief functional desc from the official pulp's doc:
"Provided by pulpcore, specifies how many repository versions will be kept for a repository. For example, if set to 1, it will keep only the most-recent repository version; the rest will be automatically deleted, together with any associated publications. Note, however, that repository versions that are currently being distributed are "protected", and cannot be removed. This can result in more versions being retained than specified by retain_repo_versions."

joey-grant:
Hey @asphalter. I saw this comment of yours and verified that the option does not exist in version `0.16`. I submitted a pull request to add this option.

https://github.com/pulp/squeezer/pull/171#issue-2430941461

PR #171, opened 2024-07-25
feat(rpm_repository): add retain_repo_versions parameter

joey-grant:
This change adds `retain_repo_versions` parameter option to `rpm_repository` module.

### DIFF
--- a/plugins/modules/rpm_repository.py
+++ b/plugins/modules/rpm_repository.py
@@ -38,7 +38,12 @@ options:
     version_added: "0.0.16"
   retain_package_versions:
     description:
-      - Max number of latest versions of each package to keep (optional)
+      - Max number of latest versions of each package to keep
+    type: int
+    version_added: "0.0.16"
+  retain_repo_versions:
+    description:
+      - Max number of repository versions to keep
     type: int
     version_added: "0.0.16"
 extends_documentation_fragment:
@@ -103,7 +108,13 @@ from ansible_collections.pulp.squeezer.plugins.module_utils.pulp import (
     PulpRpmRepository,
 )
 
-DESIRED_KEYS = {"autopublish", "remote", "repo_config", "retain_package_versions"}
+DESIRED_KEYS = {
+    "autopublish",
+    "remote",
+    "repo_config",
+    "retain_package_versions",
+    "retain_repo_versions",
+}
 
 
 def main():
@@ -115,6 +126,7 @@ def main():
             "remote": {},
             "repo_config": {"type": "raw"},
             "retain_package_versions": {"type": "int"},
+            "retain_repo_versions": {"type": "int"},
         },
         required_if=[("state", "present", ["name"]), ("state", "absent", ["name"])],
     ) as module:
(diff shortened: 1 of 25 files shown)


joey-grant:
This is related to https://github.com/pulp/squeezer/issues/165

plugins/modules/rpm_repository.py
```diff
@@ -41,6 +41,10 @@
       - Max number of latest versions of each package to keep (optional)
     type: int
     version_added: "0.0.16"
+  retain_repo_versions:
+    description:
+      - Max number of repository versions to keep (optional)
```
mdellweg:
```suggestion
      - Max number of repository versions to keep
```
not being required is implied by ansible. Can you adjust line 41 too?

Also version_added may be nice.

mdellweg (commented):
How do you feel about adding tests?
These modules are about to be transfered to a new framework. Adding this parameter to the tests will ensure it's not immediately broken afterwards.
(Please do not test the retain version feature, but just that the model is able to set it on a repository.)

plugins/modules/rpm_repository.py
joey-grant: Thanks for the feedback, @mdellweg . I've made the adjustments and will submit them soon. 

Can you confirm what you would like to see regarding tests? I have produced 2 tests:

1. set `retain_repo_versions` to 5
2. update `retain_repo_versions` to 1 (i.e. don't keep old versions)

I have ran and recorded the tests--they all pass. However, I see the following now:

1. all existing `tests/fixtures/rpm_repository-*.yml` files have been modified
2. 2 new fixture files were created: `test/fixtures/rpm_repository-{22,23}.yml`

What portions of the tests/results would you like committed?

Side note:
There seems to be a limitation in which every call that includes the `retain_repo_versions` parameter updates the returned `pulp_last_updated` timestamp (no other returned values differ). This is the case even when `retain_repo_versions` is not changed from the previous call. I assume this is a bug in `pulpcore` itself; however, it does limit the ability to assert `results.changed == false` on repeated calls with the same value for `retain_repo_versions`.

mdellweg: > What portions of the tests/results would you like committed?

All of them. They are separate files for technical reasons only.

> Side note: [...]

That actually sounds like a bug on the client side. When we don't want to change the value, we must not do a patch call, and that still seems to happen.

joey-grant: Hey @mdellweg, I have changed the tests to try for idempotence and am not seeing the previous issue. I believe that I must have made an error in my original work. Please see my changes and let me know if there is anything else needed for approval.

I appreciate you allowing me to contribute to this project.

mdellweg (dismissed):
This looks good.
Thank you!

mdellweg:
Looks like you need to run "make format" once.

joey-grant:
> Looks like you need to run "make format" once.

Hey @mdellweg, I've pushed the newly formatted code.

mdellweg (approved):
Thank you for following along.

Issue #152, opened 2024-03-21
deb_remote: add support for sync_installer and sync_udebs


achevalet:
`sync_installer` and `sync_udebs` are required to sync installer files and packages:
https://docs.pulpproject.org/pulp_deb/restapi.html#tag/Remotes:-Apt/operation/remotes_deb_apt_create

Any plan to support them? It would be great. Thanks.

mdellweg:
I would be happy to see a PR. Preferably with test coverage.

joey-grant:
@mdellweg, I have submitted a pull request for this issue: https://github.com/pulp/squeezer/pull/172

Let me know if anything needs to be tweaked.

PR #172, opened 2024-07-26
feat(deb_remote): add sync option parameters

joey-grant:
Changeset adds parameter options and tests for:
- sync_installer
- sync_sources
- sync_udebs

### DIFF
--- a/plugins/modules/deb_remote.py
+++ b/plugins/modules/deb_remote.py
@@ -31,6 +31,22 @@ options:
       - immediate
       - on_demand
       - streamed
+  sync_installer:
+    description:
+      - Sync installer packages
+    type: bool
+    version_added: "0.0.16"
+  sync_sources:
+    description:
+      - Sync source packages
+    type: bool
+    version_added: "0.0.16"
+  sync_udebs:
+    description:
+      - Sync installer packages
+    type: bool
+    version_added: "0.0.16"
+
 extends_documentation_fragment:
   - pulp.squeezer.pulp
   - pulp.squeezer.pulp.entity_state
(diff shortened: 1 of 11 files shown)


joey-grant:
This PR addresses https://github.com/pulp/squeezer/issues/152

tests/playbooks/deb_remote.yaml
```diff
@@ -88,6 +88,75 @@
...
+      pulp.squeezer.deb_remote:
+        name: test_deb_remote
+        sync_installer: true
+        state: present
+      register: result
+    - name: Verify modify sync_installer
+      assert:
+        that:
+          - result.changed == true
+          - result.remote.sync_installer == true
```
mdellweg: For the sake of brevity (and testing resources...) can you move all three options into the same two module invocations?

mdellweg (commented):
Thank you for this contribution.
In order to get this merged, you will need to record the fixtures for the tests. Please take a look at:
https://github.com/pulp/squeezer?tab=readme-ov-file#running-tests-against-a-pulp-in-one-container

tests/playbooks/deb_remote.yaml
joey-grant: @mdellweg, I combined the tests and added the fixture files. I've squashed the commits and added a `fixes #152` to the bottom of the commit messages. Let me know if there is anything out of order here.

mdellweg:
I'm trying to fix the CI issues here: https://github.com/pulp/squeezer/pull/174

plugins/modules/deb_remote.py
```diff
@@ -31,6 +31,22 @@
       - immediate
       - on_demand
       - streamed
+  sync_installer:
+    description:
+      - Sync installer packages
+    type: boolean
```
mdellweg: looks like this needs to be `bool`

joey-grant: Sorry for overlooking that. I've pushed the change. Let me know if there is anything else.

mdellweg (approved):

PR #175, opened 2024-08-04
feat(rpm_distribution): add pulp_labels parameter

joey-grant:
This change adds `pulp_labels` as an optional parameter to the `rpm_distribution` module. It also adds tests and fixtures.

### DIFF
--- a/plugins/modules/rpm_distribution.py
+++ b/plugins/modules/rpm_distribution.py
@@ -45,6 +45,12 @@ options:
     type: bool
     required: false
     version_added: "0.0.16"
+  pulp_labels:
+    description:
+      - Labels consisting of key, value pairs
+    type: dict
+    required: false
+    version_added: "0.0.16"
 extends_documentation_fragment:
   - pulp.squeezer.pulp
   - pulp.squeezer.pulp.entity_state
@@ -81,6 +87,15 @@ EXAMPLES = r"""
     password: password
     name: new_rpm_distribution
     state: absent
+
+- name: Add a rpm distribution label
+  pulp.squeezer.rpm_distribution:
+    pulp_url: https://pulp.example.org
+    username: admin
+    password: password
+    name: new_rpm_distribution
+    pulp_label:
+      key1: value1
 """
 
 RETURN = r"""
@@ -112,6 +127,7 @@ def main():
             "repository": {},
             "content_guard": {},
             "generate_repo_config": {"type": "bool"},
+            "pulp_labels": {"type": "dict"},
         },
         required_if=[
             ("state", "present", ["name", "base_path"]),
@@ -125,7 +141,13 @@ def main():
         natural_key = {"name": module.params["name"]}
         desired_attributes = {
             key: module.params[key]
-            for key in ["base_path", "generate_repo_config", "publication", "repository"]
+            for key in [
+                "base_path",
+                "generate_repo_config",
+                "publication",
+                "repository",
+                "pulp_labels",
+            ]
             if module.params[key] is not None
         }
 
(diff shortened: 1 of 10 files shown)


plugins/modules/rpm_distribution.py
```diff
@@ -112,6 +127,7 @@ def main():
             "repository": {},
             "content_guard": {},
             "generate_repo_config": {"type": "bool"},
+            "pulp_labels": {"type": "dict"},
```
mdellweg, line 130: There's a way to specify this as a dictionary of strings. Let's do that.

mdellweg (changes requested):
Thank you for that addition.
I do have one improvement in mind though.

plugins/modules/rpm_distribution.py
joey-grant: Hey @mdellweg. I've looked into this and cannot determine the method you describe. Specifically, it doesn't appear that `argument_spec` allows for arbitrary keys, which `pulp_labels` does. I could potentially change `pulp_labels` to a list of dictionaries, but that seems generally unhelpful to me. Could you provide some additional guidance if you know of a better way?

[Here is some discussion I found](http://webcache.googleusercontent.com/search?q=cache%3Ahttps%3A%2F%2Fgroups.google.com%2Fg%2Fansible-devel%2Fc%2FyYRWbLH5Zc4&sca_esv=b8c8c187ff72a731&source=hp&ei=Wb-wZrnnCOer5NoPu6nUyAI&iflsig=AL9hbdgAAAAAZrDNaShwIneiR1si3UODdgHyjj-su04t&ved=0ahUKEwj59_ed592HAxXnFVkFHbsUFSkQ4dUDCBg&uact=5&oq=cache%3Ahttps%3A%2F%2Fgroups.google.com%2Fg%2Fansible-devel%2Fc%2FyYRWbLH5Zc4&gs_lp=Egdnd3Mtd2l6Ij1jYWNoZTpodHRwczovL2dyb3Vwcy5nb29nbGUuY29tL2cvYW5zaWJsZS1kZXZlbC9jL3lZUldiTEg1WmM0Mg0QABiABBixAxhGGPkBMggQLhiABBixAzIIEAAYgAQYsQMyERAuGIAEGLEDGIMBGMcBGK8BMggQABiABBixAzIXEC4YgAQYsQMYgwEYxwEYmAUYmgUYrwEyBRAAGIAEMgUQABiABDIIEC4YgAQYsQMyCBAAGIAEGLEDSNwPUABYzQtwAHgAkAEAmAGBAaAB9QSqAQMyLjS4AQPIAQD4AQH4AQKYAgagAp8FwgIREC4YgAQYsQMY0QMYgwEYxwHCAgsQLhiABBixAxiDAcICDhAuGIAEGLEDGNEDGMcBwgILEC4YgAQY0QMYxwHCAg4QABiABBixAxiDARiKBcICCxAuGIAEGMcBGK8BwgIOEC4YgAQYsQMYgwEYigXCAgoQABiABBixAxgKwgIFEC4YgATCAgsQLhiABBixAxjUApgDAJIHAzEuNaAH1VU&sclient=gws-wiz), for reference.

mdellweg: Have you tried "elements" or does that only work with "list"? I may have been wrong.

For reference: https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec

joey-grant: Unfortunately, that is only for use with the `list` type. I did give it a try just for fun, but it seems documentation does in fact match implementation on this one.

```
"msg": "Invalid type dict for option '{'key1': 'value1', 'key2': 'value2'}', elements value check is supported only with 'list' type"
```

mdellweg (approved):

PR #176, opened 2024-08-15
added gpgkey parameter to deb_remote module

losten-git:
The deb_remote module lacks a gpgkey parameter, which is crucial for me. So i tried to implement it.

At this time, once a gpgkey is set, there is no option to set it to null. I have been unable to determine how to do so. I would appreciate assistance with this matter.

### DIFF
--- a/plugins/modules/deb_remote.py
+++ b/plugins/modules/deb_remote.py
@@ -23,6 +23,10 @@ options:
     description:
       - Whitespace separated list of distributions to sync.
     type: str
+  gpgkey:
+    description:
+      - gpgkey of remote repository
+    type: str
   policy:
     description:
       - Whether downloads should be performed immediately, or lazy.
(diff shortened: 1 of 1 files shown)


mdellweg:
Thank you! The code change looks good so far.
How do you feel about augmenting the deb_remote test to also add this parameter in one of the module invocations?
Maybe this one: https://github.com/pulp/squeezer/blob/develop/tests/playbooks/deb_remote.yaml#L65
Heads up this requires rerecording the fixtures for it.

What do you mean "I have been unable to determine how to set it to null."? Did you not manage using this change or is it impossible doing so in the api? Have you tries setting it to ""? [`null` from yaml is sadly not translated unambiguously into `None` by ansible. So we need a different indicator for "clear".]

losten-git:
>Did you not manage using this change

The change is functioning as intended.

>is it impossible doing so in the api

It is possible to set the GPG key to null using the REST API. However, I discovered that this is not an option with the `pulp-cli`.

>Have you tries setting it to ""

Yes the result is:
```
 "msg": "HTTP Error 400: Bad Request b'{\"gpgkey\":[\"This field may not be blank.\"]}'"}
```
I get the same result when I try to set the variable to null. I think you're right that the problem is that Ansible can't properly translate "null." 

Will definitely try to implement tests.

mdellweg:
This module has been rewritten using pulp-glue. Can you revisit this piece of work?

PR #198, opened 2025-04-11
Support the autopublish flag for Python repos

### LINKED ISSUE #197: python_repository lacks the autopublish flag
While the Pulp CLI allows passing the `autopublish` flag, it is missing in `pulp.squeezer.python_repository`. However, the flag is available in `pulp.squeezer.rpm_repository`. Other artifact types may be affected, too.

```console
$ pulp python repository create --help
Usage: pulp python repository create [OPTIONS]

  Create a python repository.

Options:
  --description TEXT
  --remote TEXT                   Remote used for synching in the form
                                  '[[<plugin>:]<resource_type>:]<name>' or by
                                  href.
  --autopublish / --no-autopublish
  --retain-repo-versions INTEGER  Number of repository versions to keep.
  --labels TEXT                   JSON dictionary of labels to set on python
                                  repository (or @file containing a JSON
                                  dictionary)
  --name TEXT                     [required]
  --help                          Show this message and exit.
```

stephan2012:
Allow configuring the autopublish flag for Python repositories.

Closes #197

### DIFF
--- a/plugins/modules/python_repository.py
+++ b/plugins/modules/python_repository.py
@@ -95,7 +95,7 @@ def main():
         argument_spec={
             "name": {},
             "description": {},
-            "autopublish": {"type": "bool", "default": False},
+            "autopublish": {"type": "bool"},
         },
         required_if=[("state", "present", ["name"]), ("state", "absent", ["name"])],
     ) as module:
--- a/plugins/modules/python_repository.py
+++ b/plugins/modules/python_repository.py
@@ -19,6 +19,12 @@ options:
     description:
       - Description of the repository
     type: str
+  autopublish:
+    description:
+      - Publish new versions automatically.
+    required: false
+    type: bool
+    default: false
 extends_documentation_fragment:
   - pulp.squeezer.pulp.entity_state
   - pulp.squeezer.pulp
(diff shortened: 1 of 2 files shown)


plugins/modules/python_repository.py
```diff
@@ -89,13 +95,16 @@ def main():
         argument_spec={
             "name": {},
             "description": {},
+            "autopublish": {"type": "bool", "default": False},
```
mdellweg: I think the default here is actually dangerous.
Do you feel like adding this flag to a test to demonstrate not specifying it on the second invocation does not reset it?

stephan2012: You are right. Setting a default for `autopublish` could cause unwanted repo changes. I'll update my branch accordingly (probably not today).

stephan2012: The change works as intended in my Ansible play. However, I need to dig into the tests before I can add flag checking.

mdellweg: There's a description in the README. Please tell me if you need more help.

stephan2012:
Added some tests for the autopublish flag. After some hard fights with the VCR stuff, all tests passed for me.

Please let me know if these tests match your expectations.

PR #200, opened 2025-05-08
Sven vd be patch 1

SvenVD-be:
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ImportError: Installed 'pulp-glue' version '0.32.1' is not in '>=0.29.2,<0.32'.


mdellweg:
Please don't change the operator of the dependency. We rely on z streams not to introduce breaking changes.

mdellweg:
This should have been addressed elsewhere.

Issue #201, opened 2025-06-06
"mirror" option enablement for file_sync module


chtaylo2:
Would like to add "mirror" BOOLEN as an optional setting for the file_sync module. This is supported in upstream pulp but has not been added to Pulp Squeezer.  

Planning to submit a PR soon to include + associated tests.

chtaylo2:
@mdellweg - The PR checks failed due to Ubuntu 20.04 retirement. Any thing I need to do?  Thank you

mdellweg:
I'll take care of that. You may eventually need to rebase...

Keep an eye on: https://github.com/pulp/squeezer/pull/203

chtaylo2:
Thanks. I did a rebase which kicked off the checks and they finished successfully.

PR #202, opened 2025-06-07
Added mirror option for file_sync - #201

chtaylo2:


tests/playbooks/file_sync.yaml
```diff
@@ -62,6 +62,18 @@
...
+      pulp.squeezer.file_sync:
+        remote: test_file_remote
+        repository: test_file_repository
+        mirror: true
+      register: result
+    - name: Verify sync file_remote into repository mirror=true
+      assert:
+        that:
+          - result.changed == false
+          - result.repository_version is match("/pulp/api/v3/repositories/file/file/.*/versions/1/")
```
mdellweg, line 75: So, let me understand this: It actually performs a new sync, but since it is still the same upstream, even with mirror nothing will change, and that's why we are still on version 1.
(I think that is all good, because here we want to test squeezer and not Pulp. Although, technically, we only test that the module accepts the "mirror" flag.)

plugins/modules/file_sync.py
```diff
@@ -22,6 +22,13 @@
       - Name of the remote to synchronize with
     type: str
     required: false
+  mirror:
+    description:
+      - Whether to synchronise in mirror mode.
```
mdellweg: `...ise` or `...ize`?
I appreciate the fact that the British version is probably more original, but at least we should try to be consistent within the same file.

mdellweg (commented):
Would you mind rebasing so we can merge the branch with a single commit?

tests/playbooks/file_sync.yaml
chtaylo2: Yes, that's correct. I wanted to be consistent with another Squeezer mirror test with deb_sync: https://github.com/pulp/squeezer/blob/develop/tests/playbooks/deb_sync.yaml#L66-L76

plugins/modules/file_sync.py
chtaylo2: Good call. That's been fixed and squashed

tests/playbooks/file_sync.yaml
mdellweg: That's ok. I just needed to convince myself that there really is not a new version expected.

mdellweg (approved):

mdellweg:
Thank you very much!

Issue #206, opened 2025-06-23
Pulp Glue Default timeout change


chtaylo2:
Current timeout for pulp_glue is set for 10 secs. This causes multiple tasks to fail. Each module could submit one-off timeout changes, but feel this makes the most sense to fix as a central default.

Would like to change to 3600 secs

PR #207, opened 2025-06-24
Pulp Glue default timeout change #206

chtaylo2:


mdellweg (approved):
Now I wonder if 1h is sufficient for sync tasks...

Issue #231, opened 2026-03-23
deb content missing package management


chtaylo2:
Need a deb_package playbook to allow managing DEB content in Debian repos. I'll be submitting a PR with required additions.

PR #232, opened 2026-03-24
Added deb_package to manage debian packages - #231

chtaylo2:
https://github.com/pulp/squeezer/issues/231

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -19,7 +19,7 @@ authors:
   - "Chris Taylor <taylorcw@gmail.com>"
   - "Stefan Joosten <stefan@atcomputing.nl>"
   - "Maciej Markowski <maciekmm@o2.pl>"
-version: "0.4.0-dev"
+version: "0.5.0-dev"
 license_file: "LICENSE"
 readme: "README.md"
 repository: "https://github.com/pulp/squeezer"
--- a/plugins/module_utils/pulp_glue.py
+++ b/plugins/module_utils/pulp_glue.py
@@ -39,7 +39,7 @@ class SqueezerException(Exception):
     pass
 
 
-__VERSION__ = "0.4.0-dev"
+__VERSION__ = "0.5.0-dev"
 
 
 class PulpAnsibleModule(AnsibleModule):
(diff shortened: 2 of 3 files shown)


tests/playbooks/data/dummy-package.1.0.all.deb
mdellweg, line 1: I'm not quite happy to add an opaque binary blob to a repository these days.
Do you think you can reuse one of the generated packages from our fixtures server?

mdellweg (commented):
Thank you for picking this up.

I don't think addressing packages by digest (specifically to remove them from a repository) is a convenient way. Can you select packages by name version and architecture instead?

tests/playbooks/data/dummy-package.1.0.all.deb
chtaylo2: I agree, digest is a less optimal way, though I'm leveraging the same options available to rpm_package and file_content. I feel we should have a separate enhancement story that addresses this for all.  Would you agree with this?

In regards to fixtures, pulling in something like: https://fixtures.pulpproject.org/debian_update/pool/jotunheimr/g/geirrodr/geirrodr_1.0_armeb.deb, would that be sufficient?

mdellweg: Yes, and yes.

chtaylo2: Fixture changed and created a new story for package lookup. https://github.com/pulp/squeezer/issues/233

meta/runtime.yml
```diff
@@ -43,5 +44,5 @@ action_groups:
   - task
   - user
   - x509_cert_guard
-requires_ansible: '>=2.16.0,<2.20'
+requires_ansible: '>=2.16.0,<2.22'
```
mdellweg, line 47: Sorry, should have commented on this earlier.
This is unrelated. Can you move it to a separate pr so we can look at the test matrix in the ci there. It should probably be updated too.

tests/playbooks/data/dummy-package.1.0.all.deb
mdellweg: Thank you.

chtaylo2:
@mdellweg  - Can this be merged?

mdellweg:
Please rebase the commit instead of merging from devel.

PR #240, opened 2026-05-03
add support for pulp_labels key in rpm_repository

### LINKED ISSUE #239: rpm_repository lacks the pulp_labels option
The [pulp_rpm API](https://pulpproject.org/pulp_rpm/restapi/#tag/Repositories:-Rpm/operation/repositories_rpm_rpm_create) supports adding `pulp_labels` to a rpm repository on creation time.

This key seems to be missing in squeezer so far.

philnewm:
Resolves https://github.com/pulp/squeezer/issues/239

Adds the `pulp_labels` parameter to the rpm_repository module and extends the tests based on the ones from `repo_config` parameter.

Other repository modules might require this enhancement as well but for now I only need this one.


plugins/modules/rpm_repository.py
```diff
@@ -46,6 +46,11 @@
       - Max number of repository versions to keep
     type: int
     version_added: "0.0.16"
+  pulp_labels:
+    description:
+      - A JSON document or data structure assigning pulp labels
+    type: raw
+    version_added: "0.0.16"
```
philnewm: How does this `version_added` key work?

At the time of PR creation I can't know yet which version will include it.

mdellweg: Well, we'll know at the time of merge. But really there are no automated safeguards around this. At this time the first version it could enter is 0.4.0.

mdellweg:
Please take a look at the docs testing this.

philnewm:
Did read the docs on testing and adjusted PR accordingly.
Let me know if I missed something or perhaps misunderstood.

plugins/modules/rpm_repository.py
```diff
@@ -46,6 +46,11 @@
       - Max number of repository versions to keep
     type: int
     version_added: "0.0.16"
+  pulp_labels:
+    description:
+      - A JSON document or data structure assigning pulp labels
+    type: raw
```
mdellweg: labels are `dict[str,str]`.
I don't think we should make this raw.

philnewm: Works for me, initially just copied from `repo_config`.
Changed it now

mdellweg: So I may be the origin of that then. But still i think there is `to_json` and `from_json` (with the yaml, xml, ... counterparts) in ansible. We should not add two ways to represent the same thing here.

plugins/modules/rpm_repository.py
```diff
@@ -46,6 +46,11 @@
       - Max number of repository versions to keep
     type: int
     version_added: "0.0.16"
+  pulp_labels:
+    description:
+      - A JSON document or data structure assigning pulp labels
+    type: dict[str, str]
```
mdellweg: let's hope ansible-lint tells us whether this is the supposed language here. ;)

plugins/modules/rpm_repository.py
```diff
@@ -158,6 +165,12 @@ def main():
         ):
             desired_attributes["repo_config"] = json.loads(desired_attributes["repo_config"])
 
+        # Encode the pulp_labels unless its a string, then assume it is a pre-formatted JSON
+        if "pulp_labels" in desired_attributes and isinstance(
+            desired_attributes["pulp_labels"], string_types
+        ):
```
mdellweg: And now this part is superfluous, right?

tests/playbooks/rpm_repository.yaml
```diff
@@ -205,6 +205,30 @@
...
+    - name: Verify add pulp_labels (dict) to repository
+      assert:
+        that:
+          - 'result.changed == true'
+          - 'result.repository.pulp_labels == {"test_label": "1"}'
+
+    - name: Add pulp_labels (string) to repository (Validate no change)
+      pulp.squeezer.rpm_repository:
+        name: test_rpm_repository
+        pulp_labels: '{"test_label": "1"}'
```
mdellweg, line 224: Also we don't expect that to work anymore.

Maybe:
```suggestion
        pulp_labels: "{{ '{\"test_label\": \"1\"}' | from_json }}"
```

plugins/modules/rpm_repository.py
mdellweg: Actually, this looks a bit different:
https://docs.ansible.com/projects/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec

philnewm: According to this [discussion](https://groups.google.com/g/ansible-devel/c/yYRWbLH5Zc4) and the linked [issue](https://github.com/ansible/ansible/issues/74001) it seems that the argument spec does not offer any way to define the data types of dictionary elements.
So this can only be done via runtime validation in the modules code from what I understood.

plugins/modules/rpm_repository.py
philnewm: This implementation is indeed. I replaced it with an actual check for strings.

tests/playbooks/rpm_repository.yaml
philnewm: I assumed that too but the tests still made it through. It seems that Ansible does convert inputs for dict parameters through some filter already before the data arrives at the module.

mdellweg: Ohh interesting.

mdellweg (approved):

mdellweg:
Thank you!

PR #244, opened 2026-05-17
Add support for pulp_labels key to file_repository

### LINKED ISSUE #243: file_repository lacks the pulp_labels option
The [pulp_file API](https://pulpproject.org/pulp_file/restapi/#tag/Repositories:-File/operation/repositories_file_file_create) supports adding pulp_labels to a file repository on creation time.

This key seems to be missing in squeezer so far.

philnewm:
Resolves https://github.com/pulp/squeezer/issues/243

Adds the pulp_labels parameter to the file_repository module and extends the tests based on the ones from rpm_repository.

### DIFF
--- a/plugins/modules/rpm_repository.py
+++ b/plugins/modules/rpm_repository.py
@@ -169,7 +169,9 @@ def main():
         if "pulp_labels" in desired_attributes:
             labels = desired_attributes["pulp_labels"]
             if not all(isinstance(k, str) and isinstance(v, str) for k, v in labels.items()):
-                module.fail_json(msg="pulp_labels must be a dictionary with strings as keys and values")
+                module.fail_json(
+                    msg="pulp_labels must be a dictionary with strings as keys and values"
+                )
 
         module.process(natural_key, desired_attributes)
 
--- a/plugins/modules/file_repository.py
+++ b/plugins/modules/file_repository.py
@@ -23,6 +23,11 @@ options:
     description:
       - Whether to automatically create publications for new repository versions
     type: bool
+  pulp_labels:
+    description:
+      - A dictionary assigning pulp labels using string keys and values
+    type: dict
+    version_added: "0.4.0"
 extends_documentation_fragment:
   - pulp.squeezer.pulp.entity_state
   - pulp.squeezer.pulp
@@ -86,6 +91,7 @@ except ImportError:
 DESIRED_KEYS = {
     "autopublish",
     "description",
+    "pulp_labels",
 }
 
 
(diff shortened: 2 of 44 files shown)


plugins/modules/rpm_repository.py
```diff
@@ -169,7 +169,9 @@ def main():
         if "pulp_labels" in desired_attributes:
             labels = desired_attributes["pulp_labels"]
             if not all(isinstance(k, str) and isinstance(v, str) for k, v in labels.items()):
-                module.fail_json(msg="pulp_labels must be a dictionary with strings as keys and values")
+                module.fail_json(
+                    msg="pulp_labels must be a dictionary with strings as keys and values"
+                )
```
philnewm, line 174: Adjusted the rpm_repository formatting as well to make the linter happy

mdellweg (dismissed):

mdellweg:
Looks like there is a new type of vcr incompatibility with python 3.14.? Can you look into maybe recording with a different version (I don't know exactly what the compatibility claim for that format is...)?

philnewm:
Sure, I'll take a closer look at the failing python3.14 tests later today

tests/conftest.py
```diff
@@ -2,6 +2,14 @@
...
 
 import pytest
+import yaml
+from multidict import istr
+
+# serialize istr as strings instead of python objects to stay python version agnostic (caused issues between python3.12/3.14)
+yaml.add_representer(
+    istr,
+    lambda dumper, data: dumper.represent_str(str(data)),
+)
```
philnewm: Probably not the cleanest approach but as it seems the YAML tags in the cassette fixture changed between 3.11/13 and 3.14 and they seem to require different kinds of formats now

Old format (recorded on Python 3.11/3.13):
```
? !!python/object/new:multidict._multidict_py.istr
  args:
  - Accept
  state:
    __istr_identity__: Accept
```

New format (re-recorded on Python 3.14):
```
? !!python/object/apply:multidict._multidict.istr
- Accept
```

As I don't understand much of it, this change at least made the tests happy.
Let me know if you got a better idea.

mdellweg: And the old format works with 3.14 still? Then it's probably an idea just to stick to the old (compatible) format for now.

mdellweg: With the new representation it will be just
`"Accept"`
?
That does sound clean.

mdellweg (approved):

tests/conftest.py
mdellweg: I an afterthought, I tried switching the serilizer to json. And behold, it even looks cleaner.

PR #246, opened 2026-05-30
Add tests for Ansible 2.21 and allow installation up to 2.25 

### LINKED ISSUE #245: squeezer cannot be installed with ansible-core 2.21
Ansible Core 2.21 was released recently. 
pulp.squeezer cannot be installed through ansible-galxy with that version because of https://github.com/pulp/squeezer/blob/5f198e9afab84cbcb69f531b0adf66fb1a77af5f/meta/runtime.yml#L46

This has an unexpected side effect when using the ansible-lint Github action:
- the GHA for ansible-lint always uses the latest ansible-core
- pulp.squeezer:0.0.15 was the last version with no upper limit for the ansible version, just >= 2.8
- if no version for pulp-squeezer is specified in the requirements file version 0.0.15 is selected

So for squeezer the compatibility with ansible-core 2.21 should be checked und the version requirement should be modified if possible

514amir:
Resolves #245 squeezer will keep breaking if we tightly constrain the newest ansible it supports

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,10 @@ jobs:
             ansible: "stable-2.20"
             image_tag: "nightly"
             lower_bounds: true
+          - python: "3.14"
+            ansible: "stable-2.21"
+            image_tag: "nightly"
+            lower_bounds: true
     steps:
       - uses: "actions/checkout@v6"
       - uses: "actions/cache@v5"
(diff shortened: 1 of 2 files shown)


mdellweg:
Sorry, IMHO you got it the wrong way around. With the constraint, it will reject to run yes. That is a failure, but it is not "broken". Without it however, it will be broken eventually.

Having an upper bound for exactly the version that was proven to work allows me as a maintainer to fix incompatibility issues on my schedule. And I'd rather keep it that way.

PR #247, opened 2026-06-01
Update glue

mdellweg:


514amir (commented):
@mdellweg would it possible to build a release with this please?

mdellweg:
0.4.* should contain it

